PR #41, opened 2025-01-14
fixes simulation

jucheval:
- add time sorting for Poisson processes
- change Inf to typemax in simulate_ogata
- add unit tests for Poisson process with null intensity

_I wanted to make two PRs but I made a mistake and they are combined into one._


src/simulation.jl
```diff
@@ -10,11 +10,11 @@ function simulate_ogata(
     t = tmin
     while t < tmax
         B, L = ground_intensity_bound(pp, t + eps(t), h)
-        τ = B > 0 ? rand(rng, Exponential(inv(B))) : Inf
+        τ = B > 0 ? rand(rng, Exponential(inv(B))) : typemax(inv(B))
         if τ > L
             t = t + L
         elseif τ <= L
-            U = rand(rng, T)
+            U = rand(rng)
```
gdalle, line 17:
```suggestion
            U = rand(rng, T)
```
This directly samples a variable with the right type `T` instead of `Float64`

src/simulation.jl
```diff
@@ -10,11 +10,11 @@ function simulate_ogata(
     t = tmin
     while t < tmax
         B, L = ground_intensity_bound(pp, t + eps(t), h)
-        τ = B > 0 ? rand(rng, Exponential(inv(B))) : Inf
+        τ = B > 0 ? rand(rng, Exponential(inv(B))) : typemax(inv(B))
```
gdalle, line 13:
```suggestion
        τ = B > 0 ? rand(rng, Exponential(T(inv(B)))) : typemax(T)
```
We already have the right type at our disposal: it is `T`

src/simulation.jl
```diff
@@ -10,11 +10,11 @@ function simulate_ogata(
...
         if τ > L
             t = t + L
```
gdalle, line 15:
```suggestion
            t = t + T(L)
```

src/poisson/simulation.jl
```diff
@@ -3,7 +3,7 @@ function Base.rand(
 ) where {M,T<:Real}
     mark_dist = mark_distribution(pp)
     N = rand(rng, Poisson(float(ground_intensity(pp) * (tmax - tmin))))
-    times = rand(rng, Uniform(tmin, tmax), N)
+    times = sort(rand(rng, Uniform(tmin, tmax), N))
```
gdalle, line 6: Good catch!
```suggestion
    times = sort!(rand(rng, Uniform(tmin, tmax), N))
```
This would be even more efficient because it sorts in-place

gdalle (changes requested):
Hi, thank you for this first contribution! I made a few minor remarks but overall this is nearly good to go :)

gdalle:
The tests are failing for different reasons, let me fix those for you

jucheval:
FYI, I also looked at the tests. I added these lines to Project.toml in the compat section :

LinearAlgebra = "1.11"
Random = "1.11"
Statistics = "1.11"
Aqua = "0.8"
Documenter = "1.8"
ForwardDiff = "0.10"
JuliaFormatter = "1.0"
Test = "1.11"
Zygote = "0.6"

jucheval:
Test throws also a warning for multiple definition of "f" in "test/marked_poisson_process.jl" and "test/multivariate_poisson_process.jl". We could rename them by "ll_markedpp" and "ll_multipp"

src/simulation.jl
jucheval: I am not sure that type T is relevant here. If I am correct, T is the type of the time variables. However, U is more an "intensity variable" (in particular it is compared with intensity_bound which has no reason to be of type T).

src/simulation.jl
jucheval: Aren't any chance that `T(inv(B))` throws an error ?
For instance, if `T == Int`... I agree that would not be the case in practice but it is at least possible.

src/poisson/simulation.jl
jucheval: I agree !

gdalle:
> FYI, I also looked at the tests. I added these lines to Project.toml in the compat section :

Yeah there were missing compat bounds but a lot of other stuff is outdated so I'll take care of it all in a separate PR

src/simulation.jl
jucheval: I agree, but I would say that this is handled by automatic conversion of Julia, no ?
By the way, `ground_intensity_bound` should return an `L` of type `T`.

src/simulation.jl
gdalle: No because inv(2) returns a floating point number, it doesn't throw an error

src/simulation.jl
gdalle: Essentially we have to decide which number type we want to get for the times when simulating. It can either be
- the type of the temporal bounds tmin and tmax (that is the current approach, albeit imperfectly implemented)
- the type defined implicitly by the point process (for example the type of inv(lambda))
- the promotion of the two to the most generic type

It is an arbitrary design choice, we just need to pick one option. The reason I like the current behavior is because you can easily know the type ahead of time, even when you never interact with the point process.

src/simulation.jl
jucheval: My example is:
- assume `T=Int` and `B=2`.
- Then, `inv(2) = .5 ::Float64` and `T(inv(B))`, i.e. `Int(.5)` throws an error.

gdalle: Oh right. But then the code would error anyway because we would try to create an array of times with element type `Int` and then push real numbers into it. So I would say this doesn't change much

src/simulation.jl
gdalle: I understand what you mean now, and you're right, `U` is unitless here

src/simulation.jl
```diff
@@ -10,11 +10,11 @@ function simulate_ogata(
...
         elseif τ <= L
-            U = rand(rng, T)
+            U = rand(rng)
             if U < ground_intensity(pp, t + τ, h) / B
```
gdalle, line 18:
```suggestion
            if U < ground_intensity(pp, t + τ, h) / T(B)
```

src/simulation.jl
gdalle: Actually you're right, we can keep `inv(B)` here. The conversion will happen anyway if necessary and it's more elegant

src/simulation.jl
gdalle, line 18:
```suggestion
            U_max = ground_intensity(pp, t + τ, h) / B
            U = rand(rng, typeof(U_max))
            if U < U_max
```

gdalle: how about this?

jucheval: OK

gdalle (approved):
congrats on your first merged PR! looking forward to many more :)

PR #48, opened 2025-02-05
add unit tests v2

jucheval:
This time it should be OK


Project.toml
```diff
@@ -28,6 +28,7 @@ ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+Suppressor = "fd094767-a336-5f1f-9728-57cf17d0bbfb"
```
gdalle, line 31: why is this necessary? it is rarely a good idea to suppress warnings

test/history.jl
```diff
@@ -14,6 +14,8 @@ push!(h, 1.7, "d")
 
 h2 = History([2.3], ["e"], 2.0, 2.5)
 
+@test (@capture_out show(h2)) ==
+    "History{String,Float64} with 1 events on interval [2.0, 2.5)"
```
gdalle, line 18: you can probably test `string(h2)` with the same effect

test/runtests.jl
```diff
@@ -5,6 +5,8 @@ using JuliaFormatter
 using PointProcesses
 using Random
 using Statistics
+using Suppressor
+using LinearAlgebra
```
gdalle, line 9: LinearAlgebra is not among the test dependencies I think

gdalle (changes requested):
Thanks for this initiative! A few tweaks and we're good

Project.toml
jucheval: Related with conversation below

test/history.jl
jucheval: I searched the internet for a way to test `show` methods and I found this. But you are right, `string` works.

test/runtests.jl
jucheval: You are right. I corrected it.

gdalle (approved):

Issue #49, opened 2025-02-05
Roadmap for next months


jucheval:
Here are three ideas I want to implement in the next weeks or months. I plan to follow the order presented below and make one PR for each section.
@gdalle: could you give me your opinion about this plan ?

## Remove the Mark type `M` for the models

It can be inferred from `mark_dist`, no? As for now, it is used in `fit` but I think it would be more convenient to use the type of `History{M}`.

**Add:**

- Specify that `D<:Distribution`.

## Multivariate Poisson process as a marked one (related with #43)

From what I have understood of the current code, there is no difference between them. Even the `suffstat`, `fit` and `fit_map` are similar in both cases. I will keep the folder `multivariate/` and add a test file (based on the current `test/multivariate.jl` to verify that everything is similar.

**Remove:**

- the concrete type `MultivariatePoissonProcess`
- `AbstractPoissonProcess` since there will be only one concrete type corresponding to homogeneous Poisson processes.

**Add:**

- `UnivariatePoissonProcess{T}` as an alias for `MarkedPoissonProcess{R,Dirac{Nothing}}`
- `MultivariatePoissonProcess{T}` as an alias for `MarkedPoissonProcess{R,Categorical{<:Integer}}`
- lots of constructors:
  - univariate
  - multivariate
  - marked

## Add inhomogeneous Poisson process (related with #33, #44)

With respect to #44, I will first implement inhomogeneous as a different type than homogeneous.

**Add:**

- two concrete types:
  - `HomogeneousPoissonProcess` is a renaming of `MarkedPoissonProcess` above.
  - `InhomogeneousPoissonProcess`.
- one abstract type:
  - `AbstractPoissonProcess` which is a supertype of the two concrete types above.
- a common constructor:
  - `PoissonProcess` which dispatches to Homogeneous or Inhomogeneous depending on its arguments

gdalle:
> ## Remove the Mark type M for the models

I think that's a good idea.

> It can be inferred from mark_dist, no?

Not necessarily, sometimes the only way would be to get a sample. But maybe we don't need to infer it when we don't have a history.

> Specify that `D<:Distribution`

I'm not sure what we gain from that. The whole reason I started PointProcesses.jl was to be able to use point processes inside HMMs. But I also had to write a whole new [HiddenMarkovModels.jl](https://github.com/gdalle/HiddenMarkovModels.jl) because the standard at the time in Julia (HMMBase.jl) only allowed objects from Distributions.jl.
[...]

PR #50, opened 2025-02-13
Remove mark type

jucheval:
I also added a `mark_type(pp)` function which infers type from `pp.mark_distribution`. It is used in `simulate_ogata`.


jucheval (commented):
Finally, I removed the `mark_type` function and added two lines in `simulate_ogata` to infer the type of the marks. The reason is that I didn't know how to initialize the time 0 and empty history into `mark_distribution`

src/simulation.jl
```diff
@@ -4,8 +4,10 @@
 Simulate a temporal point process `pp` on interval `[tmin, tmax)` using Ogata's algorithm.
 """
 function simulate_ogata(
-    rng::AbstractRNG, pp::AbstractPointProcess{M}, tmin::T, tmax::T
-) where {M,T<:Real}
+    rng::AbstractRNG, pp::AbstractPointProcess, tmin::T, tmax::T
+) where {T<:Real}
+    h = History(; times=T[], marks=[], tmin=tmin, tmax=tmax)
+    M = typeof(rand(mark_distribution(pp, zero(T), h)))
```
gdalle, line 10: I don't think this is going to work in general, because the first `h.marks` has element type `Any`. At the moment, the package only contains very simple test cases where `mark_distribution` actually doesn't depend on `h`, so there is no way to check.
An elegant way out of this conundrum might be to say that the argument `h` is optional in functions like `mark_distribution` or `ground_intensity`. In other words, the implementation must support no `h` being passed, meaning that we compute the relevant quantity at the very beginning of a new history. What do you think?

gdalle (changes requested):
Thanks for pinging me again, it is the right approach when I fail to answer for a while :)
Overall this looks good, the only part I'm unsure about is the initialization in Ogata's algorithm.
Also, this is a breaking change so the version number needs to be incremented accordingly.

src/simulation.jl
jucheval: In my point of view, the type of the marks produced by `mark_distribution` is inherited from the point process model `pp` rather than from the type of `h.marks`. Furthermore, I would say that `mark_distribution` should throw an error if `h.marks` cannot be converted to `Vector{M}` where `M` is the type of the marks of `pp`.

gdalle: The meaning of my remark is that in a reasonable implementation of `mark_distribution`, it would _expect_ to get a concretely-typed vector of marks inside the history `h`. If that is not satisfied (as in your proposal where `h.marks isa Vector{Any}`), then all bets are off. The code may error, or it may be type-unstable.
In retrospect, I think this issue with initializing Ogata was the reason why I decided to include the mark type in the point process. Which is not so far away from you saying 

> the type of the marks produced by mark_distribution is inherited from the point process model pp rather than from the type of h.marks.

jucheval: I understand your point. And it's also related with your comment on issue #49 ,
> > It can be inferred from mark_dist, no?

>Not necessarily, sometimes the only way would be to get a sample. But maybe we don't need to infer it when we don't have a history.

Nevertheless, I will try a last comment ;-).
I think that `convert(Vector{T}, Any[])` works for any type `T` and so I would say that your concerns is settled by automatic conversion, no ?

gdalle: But which `T` would you pick?

jucheval: Oups... you are right

So I guess that the best way is indeed to keep the mark type in `pp` as you did. I will come back to this.

jucheval (commented):
> Also, this is a breaking change so the version number needs to be incremented accordingly. 

Sorry, but I am not sure to understand the convention behind version numbers. Should I update to `0.6.0` or `1.0.0` ?

gdalle:
> In fact, if I am correct, we just have to cancel this PR, right ?

Yeah, judging by our discussion I think the existing design is better than the current proposal, which doesn't preclude the existence of a better idea somewhere that we haven't thought about yet

jucheval:
I am not sure if my previous "review" has been notified so I write this comment

gdalle:
yeah I just wanted to let you close this PR yourself to avoid being impolite but I can do it too ;)

jucheval:
I meant my last comment about the fact that your proposal about an implementation without `h` may seem better than adding type `M` to the models.

gdalle:
> I meant my last comment about the fact that your proposal about an implementation without h may seem better than adding type M to the models.

Honestly I don't think so. It forces people to add another method to every single function they define, only for the initialization of some random simulation algorithm which is very slow compared to more sophisticated ones (e.g. for Hawkes processes)

jucheval:
For most popular models, like Hawkes, I think that this method will be implemented internally. And I think that the cost of adding a new method is the same as adding `M` to the types of the model, no ?

gdalle:
Honestly this choice doesn't matter much and it's a question of esthetics and convenience more than anything, which can be subjective.
Perhaps a third option, which would be a middle ground, would be to enforce that distributions provided for the mark implement something like `mark_type(mark_dist)`. We could define it by default for objects from Distributions.jl so that most users don't worry about it.

jucheval:
Okay, I like this one.
By the way, for objects from Distributions.jl, I think that `typeof(rand(dist))` corresponds more to what we want than `eltype(dist)`. For instance, `eltype(Normal{Int}(1,0))` is `Int` whereas `typeof( Normal{Int}(1,0) )` is `Float64`

gdalle:
Indeed, but actually I just remembered that the user actually doesn't supply a distribution object, they supply a distribution factory... which can only generate `mark_dist` when supplied with a previous history of events (possibly empty). So that brings us back to square one

jucheval:
You could just mention in the API that if someone wants to come up with its own model and wants to use Oagata, then it has to implement a method with empty history.

gdalle:
I'm fine with that. I'm not sure it's a big win compared with the mark types in the model, but there's no harm in experimenting and possibly walking back the changes. If you add the warning to the docs I'd say we're good to go

gdalle:
Sorry for the endless back and forth. As we work more on this they will become less necessary, but I think it's good to keep in mind that API choices can have unintended consequences.
Ultimately you'll be taking over maintenance, so for choices which are not obviously wrong I will defer to your judgement :)

src/poisson/abstract_poisson_process.jl
```diff
@@ -24,10 +24,11 @@ end
 
 ground_intensity(pp::AbstractPoissonProcess, t, h) = ground_intensity(pp)
 mark_distribution(pp::AbstractPoissonProcess, t, h) = mark_distribution(pp)
+mark_distribution(pp::AbstractPoissonProcess, t) = mark_distribution(pp) # For simulate_ogata
```
gdalle, line 27: Are there other spots where we should add this? For instance wrappers like `BoundedPointProcess`

gdalle (commented):
Last tweak and then we're good to go

gdalle (approved):
thanks!

jucheval (commented):
Thank you too.
I waas preparing the next one in parallel so it may follow today.

src/poisson/marked/marked_poisson_process.jl
```diff
@@ -1,52 +1,26 @@
...
-"""
-    MarkedPoissonProcess(λ, mark_dist::MultivariateDistribution)
-
-Construct a `MarkedPoissonProcess` with vector marks.
-"""
-function MarkedPoissonProcess(λ, mark_dist::MultivariateDistribution)
-    M = Vector{eltype(mark_dist)}
-    return MarkedPoissonProcess{M}(λ, mark_dist)
-end
-
```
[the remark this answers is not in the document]
gdalle: > All in all, it means that we can start back with my first version of this PR (that is with the function mark_type(pp) which infers the mark type from pp.mark_distribution).

You mean add syntactic sugar like `mark_type(pp) = typeof(rand(mark_distribution(pp, tmin)))`? If so, that sounds good, but users may find it surprising that this function is not instantaneous. The point of having the mark type in the type was that `mark_type(pp)` costs nothing.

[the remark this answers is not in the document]
gdalle: > how to increment the version number ? 0.6.0 or 1.0.0 ?

Julia abides by the generic rules of semantic versioning, or [semver](https://semver.org/). For versions before 1.0.0, there a slight [tweak](https://pkgdocs.julialang.org/v1/compatibility/#compat-pre-1.0) specific to Julia.

So until v1.0.0 is released (which would mean the package is stable, and it is far from it), we only increment the minor and patch releases. Every time we break a part of the API that was documented, we increment the minor release. If we only add functionality or bug fixes, we increment the patch release.
However, v0.5.0 hasn't been _released_ yet, even though it is on the main branch of the repo. We can keep it that way for some time while you experiment with more breaking changes, and then bundle them all together in one release.

PR #51, opened 2025-03-14
Multivariate Poisson as marked

jucheval:
The goal is to merge the two concrete types `MarkedPoissonProcess` and `MultivariatePoissonProcess` as a new concrete type called `PoissonProcess` (hence it more or less corresponds to the old `AbstractPoissonProcess`)

At this stage, the two implementations coexist. In particular, there is the test file `multivariate_as_marked.jl` that checks that the old implementation of multivariate Poisson is equivalent to the new one.

To close this PR, it remains to:
- remove the files `src/poisson/abstract_poisson_process.jl` and  `test/multivariate_as_marked.jl`,
- remove and the folders `poisson/multivariate` and `poisson/marked`,
- fix the TODOs (related to aliases which would be in conflict in the current version).

### DIFF
--- a/src/poisson/fit.jl
+++ b/src/poisson/fit.jl
@@ -1,43 +1,43 @@
-## Fit from sufficient stats
+## Fit MLE
 
 function StatsAPI.fit(
-    ::Type{PoissonProcess{M,R,D}}, ss::PoissonProcessStats{R1,R2,M}
-) where {R,M,D,R1,R2}
-    λ = convert(R, ss.nb_events ./ ss.duration)
+    ::Type{PoissonProcess{R,D}}, ss::PoissonProcessStats{R1,R2}
+) where {R<:Real,D,R1<:Real,R2<:Real}
+    λ = convert(R, ss.nb_events / ss.duration)
     mark_dist = fit(D, ss.marks, ss.weights)
     return PoissonProcess(λ, mark_dist)
 end
 
-function fit_map(
-    pptype::Type{PoissonProcess{M,R,D}},
-    prior::PoissonProcessPrior,
-    ss::PoissonProcessStats{R1,R2,M},
-) where {R,M,D,R1,R2}
-    (; α, β) = prior
-    posterior_nb_events = ss.nb_events .+ α 
-    ## Note that there was a -1 here but I think it was a mistake
-    posterior_duration = ss.duration + β
-    ss_posterior = PoissonProcessStats(
-        posterior_nb_events, posterior_duration, ss.marks, ss.weights
-    )
-    return fit(pptype, ss_posterior)
-end
-
-## Fit from observations
-
 function StatsAPI.fit(
-    pptype::Type{PoissonProcess{M,R,D}}, args...; kwargs...
-) where {M,R,D}
+    pptype::Type{<:PoissonProcess}, args...; kwargs...
+)
     ss = suffstats(pptype, args...; kwargs...)
     return fit(pptype, ss)
 end
 
+## Bayesian fit (only for MultivariatePoissonProcess)
+# TODO: Replace PoissonProcess{R,Categorical{R,Vector{R}}} by
+# MultivariatePoissonProcess{R}
+# when everything else is OK
+
 function fit_map(
-    pptype::Type{PoissonProcess{M,R,D}},
+    pptype::Type{PoissonProcess{R,Categorical{R,Vector{R}}}},
+    prior::PoissonProcessPrior,
+    ss::PoissonProcessStats,
+) where {R<:Real}
+    (; α, β) = prior
+    posterior_nb_events = [sum(ss.marks .== i) for i in 1:length(α)] .+ α
+    posterior_duration = ss.duration + β
+    λ = convert(Vector{R}, posterior_nb_events ./ posterior_duration)
+    return pptype(sum(λ), Categorical(λ ./ sum(λ)))
+end
+
+function fit_map(
+    pptype::Type{<:PoissonProcess{R,Categorical{R,Vector{R}}} where R<:Real},
     prior::PoissonProcessPrior,
     args...;
     kwargs...,
-) where {M,R,D}
+)
     ss = suffstats(pptype, args..., kwargs...)
     return fit_map(pptype, prior, ss)
 end
--- a/src/poisson/fit.jl
+++ b/src/poisson/fit.jl
@@ -8,9 +8,7 @@ function StatsAPI.fit(
     return PoissonProcess(λ, mark_dist)
 end
 
-function StatsAPI.fit(
-    pptype::Type{<:PoissonProcess}, args...; kwargs...
-)
+function StatsAPI.fit(pptype::Type{<:PoissonProcess}, args...; kwargs...)
     ss = suffstats(pptype, args...; kwargs...)
     return fit(pptype, ss)
 end
@@ -33,7 +31,7 @@ function fit_map(
 end
 
 function fit_map(
-    pptype::Type{<:PoissonProcess{R,Categorical{R,Vector{R}}} where R<:Real},
+    pptype::Type{<:PoissonProcess{R,Categorical{R,Vector{R}}} where {R<:Real}},
     prior::PoissonProcessPrior,
     args...;
     kwargs...,
--- a/src/poisson/fit.jl
+++ b/src/poisson/fit.jl
@@ -1,0 +1,43 @@
+## Fit from sufficient stats
+
+function StatsAPI.fit(
+    ::Type{PoissonProcess{M,R,D}}, ss::PoissonProcessStats{R1,R2,M}
+) where {R,M,D,R1,R2}
+    λ = convert(R, ss.nb_events ./ ss.duration)
+    mark_dist = fit(D, ss.marks, ss.weights)
+    return PoissonProcess(λ, mark_dist)
+end
+
+function fit_map(
+    pptype::Type{PoissonProcess{M,R,D}},
+    prior::PoissonProcessPrior,
+    ss::PoissonProcessStats{R1,R2,M},
+) where {R,M,D,R1,R2}
+    (; α, β) = prior
+    posterior_nb_events = ss.nb_events .+ α 
+    ## Note that there was a -1 here but I think it was a mistake
+    posterior_duration = ss.duration + β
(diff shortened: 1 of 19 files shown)


jucheval:
The code coverage is worse than before because of the twin implementation of MultivariatePoissonProcess

src/poisson/fit.jl
```diff
@@ -0,0 +1,41 @@
...
+# MultivariatePoissonProcess{R}
+# when everything else is OK
+
+function fit_map(
+    pptype::Type{PoissonProcess{R,Categorical{R,Vector{R}}}},
+    prior::PoissonProcessPrior,
+    ss::PoissonProcessStats,
+) where {R<:Real}
+    (; α, β) = prior
+    posterior_nb_events = [sum(ss.marks .== i) for i in 1:length(α)] .+ α
```
gdalle, line 27:
```suggestion
    posterior_nb_events = [sum(==(i), ss.marks) for i in 1:length(α)] .+ α
```
This avoids allocations

src/poisson/multivariate/fit.jl
```diff
@@ -13,7 +13,8 @@ function fit_map(
     ss::MultivariatePoissonProcessStats,
 ) where {R}
     (; λ_α, λ_β) = prior
-    posterior_nb_events = ss.nb_events .+ λ_α .- one(eltype(λ_α))
+    posterior_nb_events = ss.nb_events .+ λ_α
+    ## Note that there was a -1 above but I think it was a mistake
```
gdalle, line 17: Yeah I think you're right, good catch

src/poisson/poisson_process.jl
```diff
@@ -0,0 +1,104 @@
...
+- `mark_dist::D`: mark distribution.
+
+# Constructor
+
+    PoissonProcess(λ, mark_dist)
+"""
+struct PoissonProcess{R<:Real,D} <: AbstractPointProcess
+    λ::R
+    mark_dist::D
+    PoissonProcess{R,D}(λ::R, mark_dist::D) where {R,D} = new{R,D}(λ, mark_dist)
```
gdalle, line 18: This constructor is not needed, it is defined by default

src/poisson/poisson_process.jl
```diff
@@ -0,0 +1,104 @@
...
+            DomainError(
+                "λ = $λ", "PoissonProcess: the ground intensity λ must be non negative."
+            ),
+        )
+    return PoissonProcess{typeof(λ),typeof(mark_dist)}(λ, mark_dist)
+end
+
+function PoissonProcess(λ::Integer, mark_dist; check_args::Bool=true)
+    return PoissonProcess(float(λ), mark_dist; check_args=check_args)
+end
```
gdalle, line 44: Is this necessary?

src/poisson/poisson_process.jl
```diff
@@ -0,0 +1,104 @@
...
+PoissonProcess() = PoissonProcess(1.0)
+
+## Access
+ground_intensity(pp::PoissonProcess) = pp.λ
+mark_distribution(pp::PoissonProcess) = pp.mark_dist
+# TODO: Replace PoissonProcess{R,Categorical{R,Vector{R}}} by
+# MultivariatePoissonProcess{R}
+# when everything else is OK
+function intensity_vector(pp::PoissonProcess{R,Categorical{R,Vector{R}}}) where {R}
```
gdalle, line 69: Is this generalizable to other kinds of processes?

gdalle: Also it needs to be documented since it is public

src/poisson/poisson_process.jl
```diff
@@ -0,0 +1,104 @@
...
+## Intensity functions
+function intensity(pp::PoissonProcess, m)
+    return ground_intensity(pp) * densityof(mark_distribution(pp), m)
+end
+
+function log_intensity(pp::PoissonProcess, m)
+    return log(ground_intensity(pp)) + logdensityof(mark_distribution(pp), m)
+end
+
+### Conversions
```
gdalle, line 82: Are these necessary?

src/poisson/prior.jl
```diff
@@ -0,0 +1,22 @@
+"""
+    PoissonProcessPrior{R1,R2}
```
gdalle, line 2:
```suggestion
    MultivariatePoissonProcessPrior{R1,R2}
```

gdalle (changes requested):
Thanks, and sorry for the review! It needs a few improvements but you can probably remove duplicate code now

src/poisson/fit.jl
jucheval: Indeed

src/poisson/poisson_process.jl
jucheval: I adapted what is done for `Exponential` in `Distributions.jl`

src/poisson/poisson_process.jl
jucheval: Do you mean that it could be a function defined for some kind of "AbstractMultivariatePointProcess" ?

src/poisson/poisson_process.jl
jucheval: Once again, I adapted it from `Distributions.jl`. If I understand correctly it is needed so that the (Any, Any) constructor is not defined by default. If I remove it, then `julia` gives a warning of method redefinition at the first constructor line 31.

src/poisson/prior.jl
jucheval: You are right. I have to change it as soon as I remove the old implementation

src/poisson/poisson_process.jl
gdalle: Or at least document it for the Poisson process

src/poisson/poisson_process.jl
gdalle: Let's remove them and stay simple. If a user tries to create a Poisson process with an integer intensity and it errors, that's on them.

src/poisson/poisson_process.jl
gdalle: The design of Distributions.jl leaves plenty to be desired and is fairly old, so it shouldn't be a universal reference. Let's avoid automatic conversions for the sake of simplicity

src/poisson/poisson_process.jl
jucheval: I replaced the inner constructor by the ones that checks if the intensity is positive.

src/poisson/poisson_process.jl
jucheval: OK

src/poisson/prior.jl
jucheval: Done

src/poisson/poisson_process.jl
jucheval: I added a docstring

src/poisson/poisson_process.jl
```diff
@@ -0,0 +1,125 @@
...
+end
+
+"""
+    MultivariatePoissonProcess{R}
+
+Homogeneous multivariate temporal Poisson process with marginal intensities of type `R`.
+
+`MultivariatePoissonProcess{R}` is simply a type alias for `PoissonProcess{R,Categorical{R,Vector{R}}}`.
+"""
+const MultivariatePoissonProcess{R<:Real} = PoissonProcess{R,Categorical{R,Vector{R}}}
```
jucheval, line 56: Just realized that the automatic conversion made by `Distributions.jl` makes, for instance, that `PoissonProcess(Int[1,1]) isa MultivariatePoissonProcess{Int}` is false.
If this "problem" happens only in the case of integers intensities, we could just don't care, but it may also arises to other `Real`'s, no ?

gdalle: Distributions.jl has a lot of issues related to types. For my package [HiddenMarkovModels.jl](https://github.com/gdalle/HiddenMarkovModels.jl), I actually rolled out my own version of `Categorical`, only to avoid them. There was a project called [MeasureTheory.jl](https://github.com/JuliaMath/MeasureTheory.jl) to rebuild Distributions.jl on a saner basis, but I think it's dead at the moment.

Either we decide that type issues are not our fault if they come from Distributions.jl, or we fix them ourselves. I'm not sure which is the best approach.

jucheval: A possibility would be to write
```suggestion
const MultivariatePoissonProcess{R<:Real} = PoissonProcess{R,Categorical{T,Vector{T}}} where T
```

jucheval:
@gdalle I think that every conversation can be resolved except the last one concerning `MultivariatePoissonProcess` for which I added a suggestion.

src/poisson/poisson_process.jl
JoseKling: From what I understand from `Categorical` in `Distributions.jl`, the type parameter is actually associated with the type of the probabilities in the probability vector. So, the type must be an `AbstractFloat`, since it is a number between 0 and 1. Except, for example, in the case `Categorical([1, 0])`, this returns a `Categorical{Int64, Vector{Int64}}`.

If the intensities are integers, the type for `Categorical` will still be an `AbstratFloat`, since it will be the type of the resulting probabilities. What seems to work is
```julia
const MultivariatePoissonProcess{R<:Real} = PoissonProcess{R,Categorical{F,Vector{F}}} where {F<:AbstractFloat}
```

JoseKling: > Here is an other suggestion :
> 
> ```julia
> const MultivariatePoissonProcess{R<:Real, F} = PoissonProcess{R,Categorical{F,Vector{F}}}
> ```
> 
> Hence, this is a `DataType`.
> 
> The drawback is that you need to provide e.g. `MultivariatePoissonProcess{Float32, Float32}` instead of `MultivariatePoissonProcess{Float32}` to the `fit` function.

Don't we need to define a separate `fit` method anyway (or, at least, `suffstats`)? From what I understand, `MultivariatePoissonProcess` is for non-marked events, so when we call `fit` with some vector of histories, we cannot use the marks in the histories, we need build the marks so that each event time has a mark corresponding to which history it came from.

If that is the case, couldn't we just define the method `fit(pp::Type{<:MultivariatePoissonProcess{R}}, hists::Vector{<:History}) where {R<:Real}`, instead of using `MultivariatePoissonPoisson{R,F}`? Or have both, but the first just dispatches to the second one with `F = float(R)`?

Issue #57, opened 2025-07-10
Minor issues with `History`


JoseKling:
### 1 - Field types

Right now, `h = History([0.5], ["a"], 0, 1)` does not work, since the same type must be used in `times`, `tmin` and `tmax`.
Any reason to not make `times`, `tmin` and `tmax` all take `Float64` and let the arguments be automatically converted?

### 2 - Add checks?

The constructor should probably check if all the events are inside the interval [`tmin`, `tmax`], it `times` is sorted and if `times` and `marks` have the same length.

### 3 - Non-marked history

Probably makes sense to have a separate constructor for non-marked history `History(times, tmin, tmax)`. Methods like `event_marks` and `max_marks` should have separate implementations, maybe just returning nothing?

JoseKling:
**1 -** I am struggling to see why not to fix the types of `times`, `tmin` and `tmax` to `Float64`. That's how I see it
1. The parametrization makes the code more complex, but with no benfit.
2. In the end, whenever we need to simulate, estimate or do almost anything these values will most probably be converted to `Float64` anyway in the calculations.
3.  Why should a `History` with `times = [1, 2, 3]`, `tmin = 0` and `tmax = 4` be any different than a `History` with `times = [1.0, 2.0, 3.0]`, `tmin = 0.0` and `tmax = 4.0`? Or even than `times = [1, 2, 3]`, `tmin = 0.0`, `tmax = 4.0`? For the marks, in contrast, this distinction makes sense.
Or am I missing something?

**2 -** Not very important, but I just thought that, in theory, since `History` is mutable, any invariant we force at creation cannot be guaranteed afterwards. For example, even with the checks, it would be possible to do
```julia
h = History([1, 2, 3], ["a", "b", "c"], 0, 4)
h.tmax = -3
```
In practice, I am not sure how much of a problem this would be, but we could simply make it non-mutable. Right now it seems that only `simulate_ogata` uses `push!`, but this could be easily modified.

jucheval:
Regarding item 1/, there are some subtypes of ‘Real’ which are used by auto diff packages. I guess that it would be weird to differentiate wrt a time of an History. However automatic differentiation may be interesting for parameters of a model (like the Hawkes you added)
Surely @gdalle has a more expert opinion on this subject.

gdalle:
I'll answer tonight in more detail but type genericity is very important in Julia programming, and I'll try to explain why later! Julien has one answer but there are many others

[...]

PR #59, opened 2025-07-10
Standard Hawkes Process

JoseKling:
- src/hawkes/hawkes_process.jl
    - HawkesProcess type
    - Methods: rand, fit, time_change, ground_intensity, integrated_ground_intensity, logdensityof

- tests/hawkes.jl
    - Some unit tests for Hawkes processes

- history.jl
    - added 'event_times(h, tmin, tmax)'
    - added 'event_marks(h, tmin, tmax)'

- simulation.jl
    - added 'simulate_poisson' - for other methods to use


src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,160 @@
...
+    return HawkesProcess(μ * (N / T), ψ * β * (N / T), β * (N / T)) # Unnormalize parameters
+end
+
+function StatsAPI.fit(
+    HP::Type{HawkesProcess}, h::History; step_tol::Float64=1e-3, max_iter::Int=1000
+)
+    fit(default_rng(), HP, h; step_tol=step_tol, max_iter=max_iter)
+end
+
+function time_change(hp::HawkesProcess, h::History)
```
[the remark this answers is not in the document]
jucheval: By the way, I would say that the computation of this $\Lambda$ was the purpose of function `integrated_ground_intensity`, no ? @gdalle

src/simulation.jl
```diff
@@ -44,3 +44,11 @@ end
 function Base.rand(pp::AbstractPointProcess, args...; kwargs...)
     return rand(default_rng(), pp, args...; kwargs...)
 end
+
+# Internal function for simulating homogeneous Poisson processes
+function simulate_poisson(rng::AbstractRNG, λ, T)
```
jucheval, line 49: Why don't use `rand(rng, Poisson(lambda), 0, T)` ?

test/hawkes.jl
```diff
@@ -0,0 +1,25 @@
...
+    hp.μ +
+      ((hp.α / hp.β) * (exp(-hp.β) - exp(-hp.β * 2))) +
+      ((hp.α / hp.β) * (1 - exp(-hp.β)))
+# Simulation
+h_sim = rand(hp, 0.0, 10.0)
+@test issorted(h_sim.times)
+@test isa(h_sim, History{Nothing,Float64})
+# Estimation
+hp_est = fit(HawkesProcess, h_sim)
+@test isa(hp_est, HawkesProcess)
```
jucheval, line 25: add a test to check that the estimated parameters are close to the real ones ?

test/history.jl
```diff
@@ -7,6 +7,12 @@ h = History([0.2, 0.8, 1.1], ["a", "b", "c"], 0.0, 2.0);
 @test !has_events(h, 1.5, 2.0)
 @test min_mark(h) == "a"
 @test max_mark(h) == "c"
+@test event_times(h) == [0.2, 0.8, 1.1]
+@test event_times(h, 0.2, 0.8) == [0.2]
```
jucheval, line 11: I thought that a `History` between `tmin` and `tmax` meant a history on the closed time interval $[tmin, tmax]$, no ?

test/hawkes.jl
```diff
@@ -0,0 +1,25 @@
...
+    (hp.μ * duration(h)) +
+      (hp.α / hp.β) * sum([1 - exp(-hp.β * (h.tmax - ti)) for ti in h.times])
+@test h_transf.times ≈ [1, (2 + (1 - exp(-2)) / 2), 4 + (2 - exp(-2 * 3) - exp(-2 * 2)) / 2]
+# Integrated ground intensity
+@test integrated_ground_intensity(hp, h, h.tmin, h.tmax) ≈ h_transf.tmax
+@test integrated_ground_intensity(hp, h, 2, 3) ≈
+    hp.μ +
+      ((hp.α / hp.β) * (exp(-hp.β) - exp(-hp.β * 2))) +
+      ((hp.α / hp.β) * (1 - exp(-hp.β)))
+# Simulation
```
jucheval, line 19: add some test for `ground_intensity` ?

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,160 @@
...
+        # Integral of activation function. 'max(tmin - ti, 0)' corrects for events that occurred
+        # inside or outside the interval [tmin, tmax].
+        integral += (exp(-hp.β * max(tmin - ti, 0)) - exp(-hp.β * (tmax - ti)))
+    end
+    integral *= hp.α / hp.β
+    integral += hp.μ * (tmax - tmin) # Integral of base rate
+    return integral
+end
+
+function DensityInterface.logdensityof(hp::HawkesProcess, h)
```
jucheval, line 128: I would say that this function does not need specific methods. I mean that as soon as functions `integrated_ground_intensity` and `log_intensity` are correctly specialized, `logdensityof` is okay thanks to its generic definition in `abstract_point_process.jl`

jucheval (commented):
I did not check every line, but there are already lots of changes requested and questions.

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,160 @@
+struct HawkesProcess <: AbstractPointProcess
+    μ::Float64
+    α::Float64
+    β::Float64
+end
```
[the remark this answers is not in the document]
JoseKling: This is related to the first item in issue #57. I don't see why we should parametrize on the type here (or in the Poisson case). In my mind it does not make sense to have, for example, integer parameters here.

Also, this would again make the initialization with `HawkesProcess(1, 2, 3.0)` fail.

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,160 @@
...
+    append!(sim, sim_desc)
+    sort!(sim)
+    return History(sim, fill(nothing, length(sim)), tmin, tmax)
+end
+
+"""
+    StatsAPI.fit(rng, ::HawkesProcess, h::History; step_tol::Float64 = 1e-3, max_iter::Int = 1000)
+
+Expectation-Maximization algorithm from [E. Lewis, G. Mohler (2011)](https://api.semanticscholar.org/CorpusID:2105771)).
+The relevant calculations are in page 4, equations 6-13.
```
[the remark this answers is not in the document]
JoseKling: In my PhD my main reference was [Chen and Hall (2013)](https://doi.org/10.1239/jap/1389370096), so I stuck with their notation.

Here it makes sense to use the notation from Lewis and Mohler. I will change it.

test/history.jl
JoseKling: Not really. The method `nb_events` also uses the half-open interval.

The history $\mathcal{H}_t$ encodes all the information available up to time $t$, which cannot include time $t$ itself.

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,160 @@
...
+    μ::Float64
+    α::Float64
+    β::Float64
+end
+
+function Base.rand(rng::AbstractRNG, hp::HawkesProcess, tmin, tmax)
+    sim = tmin .+ simulate_poisson(rng, hp.μ, tmax - tmin) # Simulate Poisson process with base rate
+    sim_desc = generate_descendants(rng, sim, tmax, hp.α, hp.β) # Recursively generates descendants from first events
+    append!(sim, sim_desc)
+    sort!(sim)
```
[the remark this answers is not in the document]
jucheval: I was thinking of using Ogata on the whole process. Each time you incorporate or discr’ard a point you update the Hawkes intensity in a manner related with what you do in your fit method (line 63)

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,160 @@
...
+    ::Type{HawkesProcess},
+    h::History;
+    step_tol::Float64=1e-3,
+    max_iter::Int=1000,
+)
+    N = nb_events(h)
+    N == 0 && return HawkesProcess(0.0, 0.0, 0.0)
+    T = duration(h)
+    c1 = 0.2 + (0.6 * rand(rng))
+    c2 = 0.1 + (0.8 * rand(rng))
```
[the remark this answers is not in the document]
jucheval: My question was a little bit oriented because I have an internship student working on this EM algorithm and he came up with some ideas for educated guesses as initial guesses in the EM algorithm. 
The underlying idea is that, in the case of a short interaction delay $\omega^{-1}$, short inter event intervals should be related to replicas whereas long inter event intervals should be related to self triggering points

test/history.jl
jucheval: Yeah, of course. My bad

src/hawkes/hawkes_process.jl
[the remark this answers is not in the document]
JoseKling: The problem is only with the efficiency.

$\Lambda(t)$ depends on all events before $t$, so using it would require repeated computation the contribution of the first events. In this way we do everything in one pass.

Unless we define a method for `integrated_ground_intensity` for a `Vector`, instead of only a single $t$.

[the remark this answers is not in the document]
jucheval: You are right, sorry. 

Regarding your last point, it might make sense if ´integrated_ground_intensity(pp, h)´ computes this vector of length ´length(h) + 2´ (we add the values in tmin and tmax).

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,160 @@
...
+    norm_ts = h.times .* (N / T) # Average inter-event time equal to 1. Numerical stability
```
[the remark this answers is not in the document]
JoseKling: This is just a optimization problem. The only thing that we need to show is that, if we start with parameters $(\mu_0, \alpha_0, \omega_0)$ and one iteration of the algorithm maps it to $(\mu_1, \alpha_1, \omega_1)$, then (normalizing -> one iteration -> unnormalizing) will also map $(\mu_0, \alpha_0, \omega_0)$ to $(\mu_1, \alpha_1, \omega_1)$.

It is actually not difficult to see that by following the calculations. Let's denote the normalized versions with a superscript $*$, so $\alpha^* = \alpha \frac{T}{N}$ is the normalized version of $\alpha$.
1. $T^* = N$, $N^* = N$
2. $\mu^* = \mu \frac{T}{N}$, $\psi^* = \psi$
3. $t_i^* = t_i \frac{N}{T}$, $\lambda(t_i)^* = \lambda(t_i) \frac{T}{N}$
4. $D^* = D$, $div^* = div \frac{T}{N}$

With this, $(\mu_0^*, \alpha_0^*, \omega_0^*) \to (\mu_1^*, \alpha_1^*, \omega_1^*) = \frac{T}{N} (\mu_1, \alpha_1, \omega_1)$, and we get $(\mu_1, \alpha_1, \omega_1)$ after undoing the normalization.

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,160 @@
...
+                diffs = norm_ts[i] - norm_ts[j]
+                D_ij = (ψ * β * exp(-β * diffs)) / lambda_ts[i] # Probability that t_i is a descendant of t_j
+                D += D_ij
+                div += (diffs * D_ij)
+            end
+        end
+        # Steps 4 and 5
+        error = max(abs(μ - (1 - (D / N))), abs(ψ - (D / N)), abs(β - (D / div)))
+        n_iters += 1
+        μ, ψ, β = 1 - (D / N), D / N, D / div # Update parameters for the next iteration
```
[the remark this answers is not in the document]
JoseKling: In the paper, the first parameter is updated with $\sum p_{ii}$, which is the expected number of immigrants conditioned on the parameters. I used $N - D$ instead, that is, $N - E[number~ of~ descendants]$.

And after normalization, $T$ becomes $N$, so $\frac{\sum p_{ii}}{T} \rightarrow \frac{N - D}{N}$.

src/hawkes/hawkes_process.jl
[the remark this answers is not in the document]
JoseKling: To be honest I don't remember exactly how I came to this, but by idea was just to have some initial guess that is compatible with the number of events observed in the process. I did not consider any other aspects.

src/hawkes/hawkes_process.jl
JoseKling: Actually, looking at it again, this is the same efficiency issue as in `integrated_ground_intensity`.

To calculate the loglikelihood, we need to calculate the sum of the intensities at each event time. With this implementation we can calculate it in one pass.

src/hawkes/hawkes_process.jl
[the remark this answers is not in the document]
JoseKling: At least how I image how this would look like it would be quite complex. The current implementation is fairly straghtforward. If we have a process over the interval $[0, T]$,
1. Simulate a Poisson process with parameter $\mu$ over $[0, T]$.
2. For each simulated event $t_s$, simulate event times $\tau_s$ from a Poisson process with parameter $\int_0^{T - t_s} \alpha e^{-\beta(t)}$ over the interval $[0, 1]$, apply the inverse of the integral, so $\Lambda^{-1}(\tau_s) \in [0, T - t_s]$. The resulting event times are $t_s + \Lambda^{-1}(\tau_s) \in [t_s, T]$.
3. Repeat step 2 for all events until the simulations generate no events.

With Ogata, we would need to keep track the intensity function to generate the next event.
1. Interval $[0, T]$, $\lambda_1(t) = \mu \rightarrow t_1 \approx Exp(1/\mu)$
2. Interval $[t_{n-1}, T]$, $\lambda_n(t) = \lambda_{n-1}(t) + \alpha e^{\omega t} \rightarrow$ first event generated by Ogata's algorithm.
3. Stop when the simulated event is larger than $T$.

Now, step 2 involves recomputing the intensity function, bounds, generating events and performing checks to see if the events should be incorporated to the history.

At least that what I could come up with. Not sure it this is similar to what you were thinking.

src/simulation.jl
JoseKling: I actually use this method to generate only a vector of times, instead of a full `History`. So this would be a bit different than the `rand` methods we define for `AbstractPointProcess`. I thought this might make some simulation algorithms simpler, but perhaps this is unnecessary.

src/hawkes/hawkes_process.jl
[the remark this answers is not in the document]
jucheval: I went through the computations with a pen and paper. I agree with you.
_FYI: I think that_ $div^* = div \frac{T}{N}$ _should be replaced by_ $div^* = div \frac{N}{T}$.

src/hawkes/hawkes_process.jl
jucheval: Indeed

test/hawkes.jl
JoseKling: Not sure if this is the best approach, but, in principle, this would show us if something changes the algorithm too much.

Could make it simpler by just fixing a seed and checking if it is close to the median, if this is too much.

src/hawkes/hawkes_process.jl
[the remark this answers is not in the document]
gdalle: This seems resolved?

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,169 @@
...
+    return History(sim, fill(nothing, length(sim)), tmin, tmax)
+end
+
+"""
+    StatsAPI.fit(rng, ::HawkesProcess, h::History; step_tol::Float64 = 1e-3, max_iter::Int = 1000)
+
+Expectation-Maximization algorithm from [E. Lewis, G. Mohler (2011)](https://api.semanticscholar.org/CorpusID:2105771)).
+The relevant calculations are in page 4, equations 6-13.
+
+Let (t_1 < ... < t_N) be the event times over the interval [0, T). We use the immigrant-descendant representation,
```
gdalle, line 29: If possible, one should refrain from using LaTeX in docstrings because it renders badly in the REPL. If you have to include formulas or symbols, it's better to put them inside backticks so that they display as monospace code, and then add a page to the Documenter.jl documentation for proper LaTeX equations

gdalle: You can visualize the docstring by typing
```
?fit
```
in a Julia REPL

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,169 @@
...
+"""
+function StatsAPI.fit(
+    rng::AbstractRNG,
+    ::Type{HawkesProcess},
+    h::History;
+    step_tol::Float64=1e-4,
+    max_iter::Int=1000,
+)
+    N = nb_events(h)
+    N == 0 && return HawkesProcess(0.0, 0.0, 0.0)
```
gdalle, line 57: Note that this should return a process of the same type as the one provided to `fit`, otherwise the function will be type-unstable. Are you familiar with that concept?
So here we should put something like
```julia
return HawkesProcess(zero(T), zero(T), zero(T))
```
and parametrize `T` inside
```julia
::Type{HawkesProcess{T}}
```

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,169 @@
...
+    T = duration(h)
+    norm_ts = h.times .* (N / T) # Average inter-event time equal to 1. Numerical stability
+    n_iters = 0
+    step = step_tol + 1.0
```
gdalle, line 61: Whenever you use floating point numbers like `1.0`, you're forcing the entire computation to happen in `Float64`, even if the user provided `Float32` everywhere. This could be replaced by `one(T)` for the right choice of `T`.

What is the right choice of `T`? Probably the result of the promotion between the time type of the `History` and the parameter type of the `HawkesProcess`

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,169 @@
...
+    lambda_ts = zeros(N)
+    # Step 1 - Choose initial guess such that Λ(T) ≈ N. After normalization, T → N
+    μ = 0.2 + (0.6 * rand(rng)) # μ should not be too close to 0 or 1. μ = 1 → the base rate already accounts for all events
```
gdalle, line 64: Not type-generic, should be `T(0.2) + T(0.6) * rand(rng, T)`

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,169 @@
...
+    ψ = (1 - μ) # ψ = α / β is the integral of the activation function. Λ(t) ≈ μt + ψN_t
+    t90 = 0.5 + (1.5 * rand(rng)) # t90 is the time it takes for the activation function to decay by 90%. Set to t90 ∈ [1/2, 2]
+    ω = log(10.0) * t90 # ω is the parameter corresponding to t90.
+    while (step >= step_tol) && (n_iters < max_iter)
+        # Step 2
+        lambda_ts[1] = 0.0
+        @inbounds for i in 2:N
```
gdalle, line 71: Never use `@inbounds` unless the bounds have been provably checked beforehand.

My rule of thumb? Never use `@inbounds`, period. As you get more and more confident writing Julia code, you'll get closer to optimal performance, but right now I'd be very surprised if bounds checking took up the bulk of the runtime. Conversely, `@inbounds` introduces the potential for segmentation faults, which is usually not worth the risk.

src/hawkes/hawkes_process.jl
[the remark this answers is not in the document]
gdalle: What are the semantics of `time_change(hp, history)`?

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,169 @@
...
+    step_tol::Float64=1e-4,
+    max_iter::Int=1000,
+)
+    N = nb_events(h)
+    N == 0 && return HawkesProcess(0.0, 0.0, 0.0)
+    T = duration(h)
+    norm_ts = h.times .* (N / T) # Average inter-event time equal to 1. Numerical stability
+    n_iters = 0
+    step = step_tol + 1.0
+    lambda_ts = zeros(N)
```
gdalle, line 62: Not type-generic: should be `zeros(T, N)`

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,169 @@
...
+           (hp.μ * duration(h)) - # Integral of base rate
+           ((hp.α / hp.ω) * sum(1 .- exp.(-hp.ω .* (duration(h) .- h.times)))) # Integral of each kernel
+end
+
+# Internal function for simulating Hawkes processes
+# The first generation, gen_0, is the `immigrants`, which is a set of event times.
+# For each t_g ∈ gen_n, simulate an inhomogeneous Poisson process over the interval [t_g, T]
+# with intensity λ(t) = α exp(-ω(t - t_g)) with the inverse method.
+# gen_{n+1} is the set of all events simulated from all events in gen_n.
+# The algorithm stops when the simulation from one generation results in no further events.
```
gdalle, line 152: You can use block comments like so:
```julia
#=
First line
second line
=#
```

src/hawkes/hawkes_process.jl
```diff
@@ -0,0 +1,169 @@
...
+function generate_descendants(rng::AbstractRNG, immigrants::Vector, T, α, ω)
+    descendants = eltype(immigrants)[]
+    next_gen = immigrants
+    while ~isempty(next_gen)
```
gdalle, line 156:
```suggestion
    while !isempty(next_gen)
```

src/history.jl
```diff
@@ -32,13 +32,31 @@ Return the sorted vector of event times for `h`.
 """
 event_times(h::History) = h.times
 
+"""
+    event_times(h, tmin, tmax)
+
+Return the sorted vector of event times between `tmin` and `tmax` for `h`.
+"""
+function event_times(h::History, tmin, tmax)
+    h.times[searchsortedfirst(h.times, tmin):(searchsortedfirst(h.times, tmax) - 1)]
```
gdalle, line 41: this could be a `@view`?

test/hawkes.jl
gdalle: Julien is right, checking proximity between estimated and true parameters is an excellent way to guard against mistakes or breaking changes.
You can use [StableRNGs.jl](https://github.com/JuliaRandom/StableRNGs.jl) to get reproducible random seeds across versions and platforms, which makes text files unnecessary.

test/hawkes_times.txt
```diff
@@ -0,0 +1 @@
...49292250425998, 0.7222898663646184, 0.7230225303240918, 0.7258703766070407, 0.7329257086975677, 0.734965964575541, 0.7367052753163055, 0.7369917675462134, 0.7377952054108926, 0.7431668020456873, 0.7442335599332027, 0.7494356056840505, 0.7592464069425915, 0.76513378821372, 0.7778400193541839, 0.783570871666886, 0.7892533149681009, 0.8151907848154715, 0.8163624371624438, 0.8203512192351479, 0.8231907075189483, 0.8254450315531231, 0.8327541619684787, 0.8340446391585701, 0.8365483397626655, 0.8380870238704485, 0.8396006152087313, 0.8424109556863868, 0.8449279007257848, 0.8493882720082039, 0.8499247624654398, 0.8570201197254786, 0.8708262595580382, 0.8757516578617132, 0.8851984472410247, 0.8943124667367437, 0.9099513775965703, 0.9158323560687229, 0.916124781335173, 0.9186099384299968, 0.9207850430116301, 0.9261702419474741, 0.9266763718478324, 0.927525059408, 0.9326181870421033, 0.9332024579347786, 0.9369755618858636, 0.9371907441403706, 0.9389797249122629, 0.939953948070523, 0.9420001470511369, 0.943798275663379, 0.9466861545841257, 0.947173907438536, 0.94880346770712, 0.9500849345227314, 0.953356815578679, 0.9535829930299445, 0.9596111063743575, 0.9686141854288621, 0.9810644372158931]
```
gdalle, line 1: See above, this is not needed

gdalle (commented):
Awesome that you're getting started with new processes!
I left a few comments to try and guide you towards idiomatic Julia code, let me know what you think!

src/hawkes/hawkes_process.jl
jucheval: Careful about the notation conflict with `T = duration(h)` on line 58

src/simulation.jl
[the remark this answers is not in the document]
JoseKling: Sorry, I probably hit `u` (vim shortcut to make everything lowercase) by mistake while having the docstring selected.

test/hawkes.jl
JoseKling: Nice. Did not know about `StableRNGs.jl`.

The problem with estimated and true parameters is that there is a lot of variability in Hawkes processes, so even if we simulate a history with known parameters, it is hard to determine how close the maximum likelihood estimator is from the true parameters. Also, the estimation relies on random initial guesses, so we might get different results in different runs.

What I came up with is this. I simulated one single history from a Hawkes process with parameters (100, 100, 200) on the interval [0, 10). I ran the estimation algorithm 10.000 times on this single history. The median gives a good approximation of what the true MLE must be, so any changes we make should still be reasonably close to this. The variance might warn us if any change in the algorithm makes it too unstable.

Still, we must assume that this implementation is correct. Since I have been working on this, I am pretty confident, but it can be tested by simulating thousands of histories from a fixed set of parameters and recording the estimated parameters from each simulation. If this is correct, the distributions of the parameters should be centered on the true parameters.

JoseKling: One other thing. Even with `StableRNGs.jl`, at some point we might change the simulation algorithm, which would cause the simulated history to change and, in turn, change the reference parameters.

src/hawkes/hawkes_process.jl
JoseKling: Yes, I am familiar with it, although it is sometimes hard to spot.

From what I understand, the method is not type stable because the first return statement always returns a `HawkesProcess{Float64}`, but the second could return a `BigFloat` (or something `Float64` could promote to), and the compiler cannot know in advance which of the statements will be hit.

In my code I was fixing the types of everything to be `Float64`, so I did not have to worry about this. But it is good to have to deal with it.

test/hawkes.jl
gdalle: > Also, the estimation relies on random initial guesses, so we might get different results in different runs.

You can't get different results if you pass the same random seed for the initialization of the algorithm too

JoseKling: True, but if the implementation changes, there is no guarantee that using the same initial guess will lead to the same estimated parameters in the end, right? That's why I thought of using the median over many estimations, so it is more robust to this randomness.

JoseKling: I guess this is fixed.
I left the file with the simulated events, because, as mentioned, if anything changes in the simulation algorithm, this will mess up the test.
I was suspicious about this variability in the estimations, but finally found that I changed the default value of the `step_tol` keyword argument for some reason. Setting it back to 1e-6 instead of 1e-3 made the estimation stable.

JoseKling:
Hello.
Any problems left here? Had some issues lately but I am back.
Cheers

test/hawkes.jl
rsenne: @JoseKling I see some definite areas we can improve performance. I will edit this week. I do have a quick question though, in the code you define t90 as:

```
    t90 = T(0.5) + (T(1.5) * rand(rng, T)) # t90 is the time it takes for the activation function to decay by 90%. Set to t90 ∈ [1/2, 2]
    ω = log(T(10)) * t90 # ω is the parameter corresponding to t90.
```

Is this current parameterization correct? If I understand correctly then:

```math
g(t_{90}) = 0.1*g(0)
```
Then we could solve the following:
```math
0.1 = e^{-\omega * t_{90}} \implies ln(10) = \omega * t_{90} \implies \omega = ln(10) / t_{90}
```

So should that line be division? or am i misreading something

rsenne (approved):

gdalle:
This is your project now, fire away :)
As a rule of thumb, and I know other package authors disagree, I'd rather have an implementation than none at all, even if some aspects can be improved later on. So I hope you feel encouraged to add stuff here and make it your own. There aren't many users right now, but once it's a bit more thorough, maybe you can solicit external feedback and that will be much more useful to guide you than my old PhD memories

PR #60, opened 2025-07-21
Implemented some goodness-of-fit tests

JoseKling:
## Statistics

Implemented two versions of the Kolmogorov-Smirnov distance as a test statistic.
1. Time-rescaled **event** times against a **uniform** distribution
2. Time-rescaled **interevent** times against a **unit exponential** distribution

## Tests

Bootstrap and non-bootstrap based tests
1. Bootstrap
    1. Estimate process
    2. Calculate test statistic
    3. Simulate data with estimated process
    4. Estimate processes from each simulation
    5. Calculate test statistic for each simulation
    6. Compare test statistic from data and from simulations to get the p value

2. Non-bootstrap
Same as above, but skip step 4, so use estimation in step 1 to calculate test statistic

## Issues
- Types and methods names
Perhaps the names can be improved. If you have any suggestions...

- Type piracy
Tried implementing a `fit` method for the `Dirac` distribution (there isn't one in `Distributions.jl`), so the `fit` method works for the `UnivariatePoissonProcess`, but tests complain about type piracy. Decided to implement a `fit` method specifically for unmarked Poisson processes instead.

### DIFF
--- a/src/HypothesisTests/PPTests/bootstrap_test.jl
+++ b/src/HypothesisTests/PPTests/bootstrap_test.jl
@@ -81,11 +81,14 @@ function BootstrapTest(
     chunk_size = max(1, length(sim_stats) ÷ Threads.nthreads())
     chunks = Iterators.partition(sim_stats, chunk_size)
 
+    # Lock for accessing the master rng safely
+    l = ReentrantLock()
+
     tasks = map(chunks) do chunk
         Threads.@spawn begin
 
             # Local rng seeded deterministically from the master rng 
-            local_rng = Xoshiro(rand(rng, UInt))
+            local_rng = lock(() -> Xoshiro(rand(rng, UInt)), l)
 
             for i in eachindex(chunk)
                 # Simulates a process and uses the process estimated from
--- a/src/HypothesisTests/PPTests/monte_carlo_test.jl
+++ b/src/HypothesisTests/PPTests/monte_carlo_test.jl
@@ -82,11 +82,14 @@ function MonteCarloTest(
     chunk_size = max(1, length(sim_stats) ÷ Threads.nthreads())
     chunks = Iterators.partition(sim_stats, chunk_size)
 
+    # Lock for accessing the master rng safely
+    l = ReentrantLock()
+
     tasks = map(chunks) do chunk
         Threads.@spawn begin
 
             # Local rng seeded deterministically from the master rng 
-            local_rng = Xoshiro(rand(rng, UInt))
+            local_rng = lock(() -> Xoshiro(rand(rng, UInt)), l)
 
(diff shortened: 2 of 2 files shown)


src/HypothesisTests/PPTests/BootstrapTest.jl
```diff
@@ -0,0 +1,83 @@
...
+test = BootstrapTest(KSDistance(Exponential), HawkesProcess, history; n_sims=1000)
+p = pvalue(test)
+```
+"""
+function BootstrapTest(
+    rng::AbstractRNG,
+    S::Type{<:Statistic},
+    PP::Type{<:AbstractPointProcess},
+    h::History;
+    n_sims=1000,
```
rsenne:
```suggestion
    n_sims::Int=1000,
```

src/HypothesisTests/PPTests/NoBootstrapTest.jl
```diff
@@ -0,0 +1,101 @@
...
+test = NoBootstrapTest(KSDistance(Exponential), HawkesProcess, history; n_sims=1000)
+p = pvalue(test)
+```
+"""
+function NoBootstrapTest(
+    rng::AbstractRNG,
+    S::Type{<:Statistic},
+    pp::AbstractPointProcess,
+    h::History;
+    n_sims=1000,
```
rsenne:
```suggestion
    n_sims::Int=1000,
```

src/HypothesisTests/pp_test.jl
```diff
@@ -0,0 +1,40 @@
...
+"""
+abstract type PPTest <: StatsAPI.HypothesisTest end
+
+"""
+    pvalue(test::PPTest)
+
+Calculate the p-value of a goodness-of-fit test on a process.
+
+# Arguments
+- `bs::BootstrapTest`: the bootstrap test result object
```
rsenne: This doesn't match the argumnet defined below

JoseKling:
Incorporated the minor suggestions, but, more importantly, refactored the multi-threaded part of the tests.

I was getting an error in `test (lts)` saying that the vector `rngs = [Xoshiro(rand(rng, UInt)) for _ in 1:Threads.nthreads]` was being accessed at `nthreads + 1`. I found out that the thread ids inside the loop were not from 1 to `nthreads` anymore, but from 2 to `nthreads + 1`. 
I looked more into it and found this [Julia Language Blog post](https://julialang.org/blog/2023/07/PSA-dont-use-threadid/) and this [Discord discussion](https://discourse.julialang.org/t/psa-thread-local-state-is-no-longer-recommended-common-misconceptions-about-threadid-and-nthreads/101274). Good to keep in mind.

PR #61, opened 2025-11-06
Small changes in `History`

JoseKling:
Simple modifications in API for `History` to solve issue #57 

- Changed arguments order, set a default `marks` argument and added the `check` keyword argument.  
`History(times, tmin, tmax, marks=fill(nothing, length(times)); check=true)`

- `History` is now immutable and checks are performed on creation. Although there is no way to prevent something like `h = History([], 0, 1, []); push!(h.times, 5)`,  I guess this is fine and still allows us to keep the previous implementation of `push!`.

- `History(times, tmin, tmax)` (no marks provided) is now accepted and returns a `History` with `nothing` as marks. Also, elements of `times`, `tmin` and `tmax` do not need to have the same type.

- Changed implementation of `append!` to be the same as `push!` for a vector of numbers. There is now a function `cat` to concatenate successive event histories.

### DIFF
--- a/src/hawkes/hawkes_process.jl
+++ b/src/hawkes/hawkes_process.jl
@@ -154,12 +154,12 @@ function StatsAPI.fit(HP::Type{HawkesProcess{T}}, h::History; kwargs...) where {
 end
 
 # Type parameter for `HawkesProcess` was NOT explicitly provided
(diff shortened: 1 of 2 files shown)


jucheval (commented):
Seems OK to me (except the minor comment...)

Issue #58, opened 2025-07-10
Conflict in `Base.rand`


JoseKling:
The method `rand(::Type{<:AbstractPointProcess}, tmin, tmax)` conflicts with the method `rand(X, d::Integer, dim::Integer)` from `Random.jl` when `tmin` and `tmax` are integers.

Any way around this? Or just change the method name to `simulate`?

jucheval:
What if you define a method `rand(::Type{<:AbstractPointProcess}, tmin::Integer, tmax::Integer)` which converts `tmin` and `tmax` to `Float`'s ? I guess that this new method should get priority over the one from `Random.jl`, no ?

JoseKling:
I tested this and apparently `Random.Sampler` is a concrete type, so `PointProcessSampler` cannot be a subtype. You can even instantiate it with, for example,

```julia
Random.Sampler(Random.default_rng(), MultivariatePoissonProcess)
```

I had a quick look at the the [link](https://docs.julialang.org/en/v1/stdlib/Random/#rand-api-hook) provided by @gdalle, but I don't really see how to implement this or what the benefit is.

One question is, how would this solve the original issue?

```julia
hp = HawkesProcess(1, 2, 3)
simulation = rand(hp, 0, 100)

ERROR: MethodError: rand(::HawkesProcess{Int64}, ::Int64, ::Int64) is ambiguous.
```

If there is no change in how we call the `rand` method, wouldn't this persist?

I think using `simulate` instead of `rand` as the method's name might make more sense. I would think we sample arrival times, but we simulate processes.

rsenne:
Yeah i think this is a good approach. I can't say I've been able to figure it out either. Maybe we can open an issue to the devs of Random and ask them about our use case?

JoseKling:
Sure, that's a good idea. 

But I was having another look, and in the documentation it says that in order to subtype `Sampler`, you need to provide the type of the output, for example,

```julia
"""
    PointProcessSampler(pp, tmin, tmax)

Sampler for a point process `pp` over the time interval `[tmin, tmax)`.
"""
@kwdef struct PointProcessSampler{PP<:AbstractPointProcess, T<:Real} <: Random.Sampler{Vector{Float64}} 
    pp::PP
    tmin::T
    tmax::T
end
```

[...]

PR #62, opened 2025-11-09
Change rand to simulate

rsenne:
This PR does two things:

1.  Changes 'rand' to 'simulate' as discussed in #58 
2. Reformats the testing directory so that the 'test' folder is its own julia project. This way we can have one set of imports in  `runtests.jl` and keep the Project.toml files separate which i find leads to fewer headaches


JoseKling:
Looks good to me. Merging.

PR #65, opened 2025-11-30
Inhomogeneous point process

### LINKED ISSUE #33: inhomogeneous Poisson processes
Hi @gdalle! 

I'm interested in using this package to work with inhomogeneous Poisson processes but I didn't see a definition in the source. Is this something that would be relatively straightforward to implement? I'm happy to work on a PR for this if you think it would be a good idea, with some guidance regarding the design of the package. Thanks.

rsenne:
Start of a PR for adding IPP models. Should resolve #33. I think this is a reasonably good start but before continuing, I think I'd like an extra set of eyes to make sure everyone is happy with the current structure. We can almsot certainly do better on the numerical integration so am open to suggestions. I also feel we may want to expand the AutoDiff support (maybe via [DifferentationInterface](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/stable/)).

This pull request introduces support for inhomogeneous Poisson processes (IPPs) to the point process modeling library. It adds new intensity function types, a flexible process type for IPPs, and comprehensive fitting methods using maximum likelihood estimation (MLE) with automatic differentiation. These changes enable modeling, simulation, and fitting of temporal Poisson processes with non-constant, parametric, or piecewise intensity functions.

### Inhomogeneous Poisson Process Support

* Added the new `InhomogeneousPoissonProcess` type, representing Poisson processes with time-varying intensity and mark distributions, along with interface methods and documentation.
* Exported `InhomogeneousPoissonProcess` and several new intensity function types (`PolynomialIntensity`, `ExponentialIntensity`, `SinusoidalIntensity`, `PiecewiseConstantIntensity`, `LinearCovariateIntensity`) to the main package API.

### Intensity Function Infrastructure

* Implemented parametric intensity function types for IPPs, including polynomial, exponential, sinusoidal, piecewise constant, and linear covariate forms, each with constructors, validation, and callable behavior.

### Fitting and Estimation

* Added MLE-based fitting methods for IPPs, supporting automatic differentiation and numerical optimization via Optim.jl. Includes specialized fitting routines for each parametric intensity type and support for fitting from multiple event histories.

### Integration and Simulation

* Integrated new IPP modules and intensity functions into the package, making them available for use and simulation.
* Added dependency on the Optim.jl package for numerical optimization in fitting routines.

### DIFF
--- a/src/poisson/inhomogeneous/intensity_methods.jl
+++ b/src/poisson/inhomogeneous/intensity_methods.jl
@@ -57,11 +57,10 @@ function intensity_bound(
     t::T,
     h::History;
     lookahead_factor::Real=1/100,
-    min_lookahead::T=T(1e-3),
     n_samples::Int=10 * length(f.coefficients),
 ) where {R,T}
     dur = duration(h)             # or max_time(h) - min_time(h)
-    lookahead = max(T(lookahead_factor * dur), min_lookahead)
+    lookahead = T(lookahead_factor * dur)
     return intensity_bound(f, t; lookahead=lookahead, n_samples=n_samples)
 end
 
@@ -100,14 +99,10 @@ function intensity_bound(f::ExponentialIntensity{R}, t::T; lookahead::T=one(T)) 
 end
 
 function intensity_bound(
-    f::ExponentialIntensity{R},
-    t::T,
-    h::History;
-    lookahead_factor::Real=1/100,
-    min_lookahead::T=T(1e-3),
+    f::ExponentialIntensity{R}, t::T, h::History; lookahead_factor::Real=1/100
 ) where {R,T}
     dur = duration(h)
-    lookahead = max(T(lookahead_factor * dur), min_lookahead)
+    lookahead = T(lookahead_factor * dur)
     return intensity_bound(f, t; lookahead=lookahead)
 end
 
@@ -230,11 +225,10 @@ function intensity_bound(
     t::T,
     h::History;
     lookahead_factor::Real=1/100,
-    min_lookahead::T=T(1e-3),
     n_samples::Int=100,
 ) where {R,T}
     dur = duration(h)
-    lookahead = max(T(lookahead_factor * dur), min_lookahead)
+    lookahead = T(lookahead_factor * dur)
     return intensity_bound(f, t; lookahead=lookahead, n_samples=n_samples)
 end
 
@@ -262,14 +256,9 @@ function intensity_bound(f::F, t::T; lookahead::T=one(T), n_samples::Int=100) wh
 end
 
 function intensity_bound(
-    f::F,
-    t::T,
-    h::History;
-    lookahead_factor::Real=1/100,
-    min_lookahead::T=T(1e-3),
-    n_samples::Int=100,
+    f::F, t::T, h::History; lookahead_factor::Real=1/100, n_samples::Int=100
 ) where {F,T}
     dur = duration(h)
-    lookahead = max(T(lookahead_factor * dur), min_lookahead)
+    lookahead = T(lookahead_factor * dur)
     return intensity_bound(f, t; lookahead=lookahead, n_samples=n_samples)
 end


src/poisson/inhomogeneous/intensity_methods.jl
```diff
@@ -0,0 +1,236 @@
...
+        end
+        return integral
+    end
+end
+
+"""
+Upper bound for polynomial intensity over an interval.
+
+For polynomials, we sample densely and add a margin.
+"""
```
JoseKling: I think this method should take into account the duration of the history. It should be possible to have durations smaller or much larger than 1.
Perhaps define a lookahead of `duration(h) / 1000` or something like that.

src/poisson/inhomogeneous/fit.jl
```diff
@@ -0,0 +1,371 @@
...
+
+    # Optimize using Optim.jl with automatic differentiation
+    result = optimize(objective, initial_params, optimizer; autodiff=autodiff)
+
+    # Extract optimal parameters
+    optimal_params = Optim.minimizer(result)
+
+    # Build final intensity function
+    intensity_func = param_to_intensity(optimal_params)
+
```
JoseKling: where is the `intensity_to_params` being used?

src/poisson/inhomogeneous/intensity_methods.jl
```diff
@@ -0,0 +1,236 @@
...
+    if f.link === :identity
+        # Analytical integral for polynomial
+        result = zero(promote_type(eltype(f.coefficients), typeof(a), typeof(b)))
+        for (i, coef) in enumerate(f.coefficients)
+            power = i
+            result += coef * (b^power - a^power) / power
+        end
+        return result
+    else
+        # Numerical integration for log link
```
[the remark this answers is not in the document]
rsenne: Maybe we can add them as keyword arguments and let the user decide and otherwise use some reasonable initial values?

rsenne:
Okay i re-added the `PiecewiseConstant` fit function and all tests pass again with your new interface for fitting. As for point 2, I agree that this isn't a perfect solution, but I think the alternative of not having to would be more prone to errors given the optimizer could wander into spaces where the intensity function has negative values. I think if our goal is to reasonably avoid all errors, then i think the solution is to make sure all `Parametric` models are fit as:

```math
Poisson(\lambda_t(t) = \exp^{f(\theta)})
```
and not

```math
Poisson(\lambda_t(t) = f(\theta))
```
I myself actually prefer this as if a user has a reason to fit some other model it leaves these concerns up to them.

src/poisson/inhomogeneous/fit.jl
```diff
@@ -0,0 +1,134 @@
...
+    h::History,
+    init_params;
+    optimizer=LBFGS(),
+    autodiff=:forward,
+    integration_config=IntegrationConfig(),
+    intensity_kwargs...,
+)
+    if isempty(h.times)
+        return InhomogeneousPoissonProcess(t -> 0.0, Dirac(nothing))
+    end
```
JoseKling: Besides the point that I returned the wrong type, I guess this is where you used the `intensity_to_params` in some previous commit. So we could return a constant 0 intensity of the same type.

An empty history should not cause any problems to the optimizer, it will just try to minimize the integral of the intensity. So we could just get rid of this.

src/poisson/inhomogeneous/inhomogeneous_poisson_process.jl
```diff
@@ -0,0 +1,109 @@
...
+"""
+    ground_intensity_bound(pp::InhomogeneousPoissonProcess, t, h)
+
+Compute a local upper bound on the ground intensity.
+
+This delegates to `intensity_bound` defined for specific intensity function types
+in intensity_methods.jl. Analytical bounds are used when available.
+"""
+function ground_intensity_bound(pp::InhomogeneousPoissonProcess, t::T, h) where {T}
+    return intensity_bound(pp.intensity_function, t)
```
JoseKling: `h` must be passed to `intensity_bound`

JoseKling (commented):
> I think if our goal is to reasonably avoid all errors, then i think the solution is to make sure all `Parametric` models are fit as:
>                                 $$Poisson(\lambda t(t) = expf(\theta))$$

I think this is reasonable, but I also think this is not an obvious behavior an user would expect. If I created an intensity, I would not expect it to be transformed in any way without me choosing it.

An alternative would be to set up something similar to the `IntegrationConfig()`. We could have a `OptimizationConfig()` for the to choose the solver, the constrains, or even add gradients. Not sure how exactly this would look like.

Anyway, I am fine leaving it as is or with this transformation you proposed for now. What you did is already a lot of functionality added, and we can improve this in the future.

src/poisson/inhomogeneous/intensity_methods.jl
```diff
@@ -0,0 +1,225 @@
...
+#=
+Upper bound for sinusoidal intensity.
+
+Maximum is a + |b| (when sin = 1 if b > 0, or sin = -1 if b < 0).
+=#
+function intensity_bound(f::SinusoidalIntensity{R}, t::T, h::History) where {R,T}
+    B = f.a + abs(f.b)
+    L = typemax(T)  # Bound holds for all time
+    return (B, L)
+end
```
JoseKling: Here I understand not using `h`. It could be more efficient to have a better bound, but it is fine.

src/poisson/inhomogeneous/intensity_methods.jl
```diff
@@ -0,0 +1,225 @@
...
+
+# Numerical bound for arbitrary callable intensity functions.
+function intensity_bound(f::F, t::T, h::History) where {F,T}
+    lookahead = T(1.0)
+    n_samples = 100
+    ts = range(t, t + lookahead; length=n_samples)
+    max_intensity = maximum(f(ti) for ti in ts)
+    return (max_intensity * T(1.1), lookahead)
+end
```
JoseKling: Same problem here. Just use a lookahead of `duration(h) / 100` or something like that.

rsenne:
Should be all done @JoseKling let me know if anything else pops up

src/poisson/inhomogeneous/intensity_methods.jl
```diff
@@ -0,0 +1,275 @@
...
+    f::ExponentialIntensity{R},
+    t::T,
+    h::History;
+    lookahead_factor::Real=1/100,
+    min_lookahead::T=T(1e-3),
+) where {R,T}
+    dur = duration(h)
+    lookahead = max(T(lookahead_factor * dur), min_lookahead)
+    return intensity_bound(f, t; lookahead=lookahead)
+end
```
JoseKling:
```suggestion
function intensity_bound(
    f::ExponentialIntensity{R},
    t::T,
    h::History;
    lookahead_factor::Real=1/100,
) where {R,T}
    dur = duration(h)
    lookahead = T(lookahead_factor * dur)
    return intensity_bound(f, t; lookahead=lookahead)
end
```

src/poisson/inhomogeneous/intensity_methods.jl
```diff
@@ -0,0 +1,275 @@
...
+    t::T,
+    h::History;
+    lookahead_factor::Real=1/100,
+    min_lookahead::T=T(1e-3),
+    n_samples::Int=100,
+) where {R,T}
+    dur = duration(h)
+    lookahead = max(T(lookahead_factor * dur), min_lookahead)
+    return intensity_bound(f, t; lookahead=lookahead, n_samples=n_samples)
+end
```
JoseKling:
```suggestion
function intensity_bound(
    f::LinearCovariateIntensity{R},
    t::T,
    h::History;
    lookahead_factor::Real=1/100,
    n_samples::Int=100,
) where {R,T}
    dur = duration(h)
    lookahead = T(lookahead_factor * dur)
    return intensity_bound(f, t; lookahead=lookahead, n_samples=n_samples)
end
```

Issue #73, opened 2025-12-19
Representation of multivariate processes


JoseKling:
Following PR #51, multivariate processes are represented as marked processes with i.i.d. integer marks, each mark representing one marginal process. This is fine for Poisson processes, since the marks do not influence the intensity of the process.

For Hawkes processes, however, marks may influence the intensity, so a univariate Hawkes process with integer marks is not equivalent to a multivariate Hawkes process. Currently I am treating any integer marks as really meaning the marginal process an event occurs, but I think this is not great. 

This was shortly discussed in #43. I am not currently considering processes with vector-valued marks, so the options I see are:

1. Define that this package deals only with univariate processes, but because of the structure of the Poisson process, we can interpret this univariate process as a multivariate one.

2.  Fields `times` and `marks` of `History` hold vectors of vectors, instead of vectors. 

3. Add another field that holds a vector containing the marginal process corresponding to each event. If this field contains a vector filled with `nothing`, than it is a univariate process.

4. Have multiple subtypes of `History`, which may or may not have the fields `marks` and `marginal_process` (or whatever name).
 
I would not consider (1), since I started working on mutually exciting Hawkes processes. Between (2) and (3), I guess (2) is a more straightforward way to represent multivariate processes, but (3) has the benefit that the event times are already ordered, which is needed for pretty much every method related with multivariate Hawkes processes. (4) is possibly the most efficient (?), but a similar functionality can be obtained using aliases.

Thoughts? Other alternatives?

JoseKling:
What I came up with, which should not change the interface too much, so we can still use what we have with small modifications is this (without the complete implementation of the constructors, so it is not too long)

```julia
struct History{N,T<:Real,M,D} # N is the number of dimensions, D can be Int or Nothing
    times::Vector{T}
    tmin::T
    tmax::T
    marks::Vector{M}
    dims::Vector{D}     # In which of the 'N' marginal process the event occurs

    # Add optional arguments check
    # Inner constructors are the standard for the methods to create a `History`
    function History{N}(times::Vector{<:Real}, tmin::Real, tmax::Real, marks::Vector, dims::Vector{<:Int}) where {N}
        new{N,T,M,D}(times, tmin, tmax, marks, dims)
    end

    function History{1}(times::Vector{<:Real}, tmin::Real, tmax::Real, marks::Vector, ::Any)
        new{1,T,M,Nothing}(times, tmin, tmax, marks, dims)
    end
end

[...]

PR #77, opened 2026-01-14
Multivariate `History`

JoseKling:
Added the parameter `N` to history, indicating the number of marginal processes in the multivariate point process.  Added methods to access the marginal histories and counts.

Virtually no change in implementation was necessary for univariate processes to keep working as before.


jucheval:
Edit: I just realized that the discussion about this PR is in Issue #73

PR #84, opened 2026-05-11
Fixes #82 and #83

### LINKED ISSUE #82: `time_change` should integrate cumulative inter-arrival segments
`time_change(h, pp::InhomogeneousPoissonProcess)` currently computes the compensator at each event time and at tmax via independent quadrature calls. Independent calls have independent ulp-level rounding errors, so the transformed times can come out of order or slightly past `new_tmax`. Switching to a cumulative integration over inter-arrival segments removes both failure modes by construction with no change in cost.

### LINKED ISSUE #83: Should `History` be a validator solely, or should it be a fixer?
See discussion on #78.

This would be a breaking change, but we aren't at major version 1 yet, so allowed under semantic versioning. If we decide to go this route, we should do a deprecation cycle anyway until v1.0.0 lands imo

rsenne:
This PR switches the default behavior of `History`, `push!`, and `append!` to no longer coerce the time of arrivals to be sorted.

This in theory will prevent two pathological cases where 1.) a user has data issues they don't know about or 2.) or they merely havent sorted their data. Ideally, this removes the layer of reasoning about the user and their data. 

We however, still retain the ability to set `check_args=false` in the event minor numerical errors occur where we may expect them in th ebackend and handle them ourselves. 

I also tacked on the propsoed chackge to time_change using `cumsum` which as better design.

Also bumps the version of the project. Resolves #82 and resolves #83.

### DIFF
--- a/src/univariate/poisson/inhomogeneous/inhomogeneous_poisson_process.jl
+++ b/src/univariate/poisson/inhomogeneous/inhomogeneous_poisson_process.jl
@@ -122,18 +122,21 @@ The transformed event times form a unit-rate (homogeneous) Poisson process on
 function time_change(h::History, pp::InhomogeneousPoissonProcess)
     f = pp.intensity_function
     config = pp.integration_config
-    new_tmax = integrated_intensity(f, h.tmin, h.tmax, config)
-    T = typeof(new_tmax)
-    new_times = T[integrated_intensity(f, h.tmin, t, config) for t in h.times]
-    # `new_tmax` and each `new_times[i]` come from independent quadrature calls,
-    # so solver / floating-point error can both swap adjacent transformed
-    # events and let one of them slightly exceed `new_tmax`. Widen `new_tmax`
-    # to the observed maximum (with an `eps` cushion) so no event is dropped
-    # at the boundary; small order drift is then handled by the History
-    # constructor's automatic `sortperm` pass under `check_args=true`.
-    if !isempty(new_times)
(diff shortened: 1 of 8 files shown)


src/univariate/poisson/inhomogeneous/inhomogeneous_poisson_process.jl
```diff
@@ -122,18 +122,21 @@ The transformed event times form a unit-rate (homogeneous) Poisson process on
...
+    adaptive quadrature returns a tiny negative result on a non-negative
+    integrand.
+    =#
+    edges = vcat(h.tmin, h.times, h.tmax)
+    deltas = [
+        integrated_intensity(f, edges[i], edges[i + 1], config) for
+        i in 1:(length(edges) - 1)
+    ]
+    T = eltype(deltas)
+    deltas .= max.(zero(T), deltas)
```
JoseKling: This actually does not guard against errors. If any delta is 0, then we have repeated events.
I guess we should use `max(eps(T), deltas)` instead. A warning could be useful, maybe?

JoseKling (changes requested):

src/univariate/poisson/inhomogeneous/inhomogeneous_poisson_process.jl
rsenne: Good catch. I went for a middle ground here. The new behavior throws a `DomainError` but I added stress tests which seems to suggest such an event occurring is rare. If the error happens to a user then I would imagine they'd have a very poorly numerically behaved integral

JoseKling (approved):

PR #85, opened 2026-05-17
History fixes 

### LINKED ISSUE #82: `time_change` should integrate cumulative inter-arrival segments

### LINKED ISSUE #83: Should `History` be a validator solely, or should it be a fixer?

rsenne:
I realized i opened up the PR on the wrong branch....so this is now the real PR lmao.

Resolves #82 and resolves #83


rsenne:
hey @JoseKling could you approve this? I mistakenly didn't setup the PR to go into main so this should be the same code but now applied to the correct branch

JoseKling (approved):

PR #88, opened 2026-05-19
Implemented interface for mark distribution

### LINKED ISSUE #80: Struct for mark distributions
The current implementations only account for mark distributions that are independent from the event times. This is mainly focused on using `Distributions.jl` as mark distributions.  
There are some obvious limitations to that and it is also a bit cumbersome at times. Some issues:
- It is not straightforward to use a `PoissonProcess` with a mark distribution that depends on time or the history
- Using the `Dirac(nothing)` for unmarked processes is annoying and not intuitive. It has no `logdensityof` method, which breaks some methods.

My idea is to add a `AbstractMarkDistribution` that can be easily extended and implements the necessary interface independently of the process. We could have things like

```julia
abstract type AbstractMarkDistribution end

sample_mark(md::AbstractMarkDistribution, h, t) = rand(md.mark_dist)
DensityInterface.densityof(md::FixedDistribution, h, t, m) = densityof(mark_distribution(md.mark_dist, h, t), m)

# Replaces `Dirac(nothing)` 
struct NoMarks <: AbstractMarkDistribution end

mark_distribution(md::NoMarks, t , h) = Dirac(nothing)
DensityInterface.densityof(md::NoMarks, h, t, nothing) = 1.0 # Fix density issue

# Works with distributions from Distribution.jl
struct FixedDistribution{D} <: AbstractMarkDistribution 
    mark_dist::D
end

mark_distribution(m::FixedDistribution, h, t) = m.mark_dist()

# Easily extend to other cases
struct NonStationaryNormalDistribution{R<:Real} <: AbstractMarkDistribution
    a::R
    b::R
end
mark_distribution(md::NonStationaryNormalDistribution, t , h) = Normal( a * t + b)
```

would require a bit more thought on the interface, though.

JoseKling:
The interface is implemented. Maybe the names can be improved.

What I implemented:
- An `AbstractMarkDistribution` type for mark distributions different from the ones in `Distributions.jl`.
- `PointProcessMarkDistribution` alias for `Union{AbstractMarkDistribution,Distribution}`.
- The following methods as part of the common interface.
  - `mark_distribution` -> The distribution of marks at some time `t` and after history `h`.
  - `sample_mark` -> Sample from `mark_distribution`. Standard implementation assumes the result of `mark_distribution` has the method `rand`.
  - `densityof` -> for mark `m` at time `t` after history `h`. Standard implementation assumes result of `mark_distribution` has method `densityof`.
  - `fit` -> Fit the distribution to a history `h`. For processes whose event times are independent of the marks.
  - `eltype` -> Used to initialize empty histories. Standard implementation assumes result of `mark_distribution` has method `eltype`.
- Added the `CustomProcesses` page to the docs.


JoseKling:
In `src/univariate/poisson/poisson_process.jl` I deleted the methods
- `ground_intensity(pp)`, `ground_intensity(pp, t)`
- `intensity(pp, m)`, `intensity(pp, t, m)` and `log_intensity`
- `mark_distribution(pp)`, `mark_distrbution(pp, t)`
- 
One reason is they are conceptually incorrect, the intensity is defined as dependent on the history. `ground_intensity(pp)` is actually the parameter $\lambda$, so it should be accessed with `pp.λ`.    
Second reason is that now that we are supporting custom mark distributions, `intensity` and `mark_distribution` cannot be assumed to not depend on `t` or `h`.  
Lastly, we just unify the interface.  

Should we raise a deprecation error or something like that?

rsenne:
Hi @JoseKling I'll try and take a look at this later in the week

src/history.jl
```diff
@@ -158,6 +162,17 @@ function History(; times, tmin, tmax, marks=nothing, dims=nothing, check_args=tr
...
 end
 
+function History(tmin::R1, tmax::R2, N::Int=1) where {R1<:Real,R2<:Real}
+    R = promote_type(R1, R2)
+    History(R[], tmin, tmax, [], [], N)
+end
+
+function History(h::History, d::Int)
+    times = event_times(h, d)
+    marks = event_times(h, d)
```
rsenne: Should this be `event_marks`?

src/history.jl
```diff
@@ -363,7 +378,7 @@ inner loops.
 function Base.push!(h::History, t::Real, m=nothing, d=nothing; check_args=true)
     if check_args
         @assert h.tmin <= t < h.tmax
-        @assert (length(h) == 0) || (h.times[end] <= t)
+        @assert (length(h) == 0) || (h.times[end] < t)
```
rsenne: The docsting and code seem to be inconsistent. Also, will this prevent a multivariate process from having event times occur simulataneously in multiple dimensions?

src/univariate/poisson/simulation.jl
```diff
@@ -1,5 +1,18 @@
...
+#     h = History(T[], tmin, tmax, eltype(pp.mark_dist)[])
+#     inter_dist = Exponential(inv(pp.λ))
+#     t = rand(rng, inter_dist)
+#     while t < tmax
+#         m = sample_mark(pp.mark_dist, t, h)
+#         push!(h, t, m; check_args=False)
+#         t = rand(rng, inter_dist)
+#     end
+#     return h
+# end
```
rsenne:
```suggestion

```

src/mark_distributions.jl
```diff
@@ -0,0 +1,69 @@
...
+function Base.eltype(md::AbstractMarkDistribution)
+    eltype(mark_distribution(md, 0.0, History(0.0, 1.0)))
+end
+
+"The likelihood of a mark `m` occurring in an event at time `t` after history `h`"
+function DensityInterface.densityof(md::PointProcessMarkDistribution, t, h::History, m)
+    densityof(mark_distribution(md, t, h), m)
+end
+
+StatsAPI.fit
```
rsenne:
```suggestion
```

src/PointProcesses.jl
```diff
@@ -38,8 +39,13 @@ export min_time, max_time, min_mark, max_mark
 export nb_events, has_events, duration, ndims
 export time_change, split_into_chunks
 
+## Mark Distributions
+export AbstractMarkDistribution, PointProcessMarkDistribution, NoMarks
+export sample_mark
+
 ## Point processes
 
+export PointProcessMarkDistribution, AbstractMarkDistribution, NoMarks
```
rsenne:
```suggestion
```

src/abstract_point_process.jl
```diff
@@ -78,7 +80,7 @@ For multivariate processes, it returns a vector of log intensities for each dime
 log_intensity(pp, h, t, d) computes the log intensity for a multivariate process `pp` at dimension `d`.
 """
 function log_intensity(pp::AbstractPointProcess, m, t, h)
-    return log(ground_intensity(pp, t, h)) + logdensityof(mark_distribution(pp, t, h), m)
+    return log(intensity(pp, m, t, h))
 end
```
rsenne, line 84: These generic methods reach directly into a pp.mark_dist field, but that field isn't part of the AbstractPointProcess standard API. It happens to exist in PoissonProcess and IPP but not in Hawkes. There it would throw an error. 

Maybe we could make this a standard part of the interface. Also, we could make this an accesor e.g.,

```julia
"""
    mark_dist(pp)

Return the mark distribution object of `pp`. Subtypes either store it in a
`mark_dist` field (the default) or override this method.
"""
mark_dist(pp::AbstractPointProcess) = pp.mark_dist
```

Then it could be overriden if needed?

src/simulation.jl
```diff
@@ -22,7 +22,7 @@ To infer the type of the marks, the implementation assumes that there is method
 function simulate_ogata(
     rng::AbstractRNG, pp::AbstractPointProcess, tmin::T, tmax::T
 ) where {T<:Real}
-    M = typeof(rand(mark_distribution(pp, tmin)))
+    M = eltype(pp.mark_dist)
```
rsenne, line 22: I think it would be good to make `eltype` a required method for any AbstractMarkDistribution

test/multivariate_poisson_process.jl
```diff
@@ -67,18 +67,18 @@ end
...
-    h2 = History(fill(rand(100) .* 1000, 3), 0, 1000, fill(rand(100), 3))
+    h1 = simulate(rng, pp2, 0.0, 1000.0)
+    h2 = History(fill(rand(100) .* 1000, 3), 0.0, 1000.0)
 
-    pp_est1 = fit(fill(PoissonProcess{Float64,Normal}, 3), h1)
-    pp_est2 = fit(fill(PoissonProcess{Float64,Normal}, 3), h2)
+    pp_est1 = fit(fill(PoissonProcess{Float64,NoMarks}, 3), h1)
+    pp_est2 = fit(fill(PoissonProcess{Float64,NoMarks}, 3), h2)
     λ_est1 = [pp_est1.processes[d].λ for d in 1:ndims(pp_est1)]
     λ_est2 = [pp_est2.processes[d].λ for d in 1:ndims(pp_est2)]
```
rsenne, line 76: this used to test a marked case. Why switch ot just NoMarks? We should test both no?

docs/src/examples/CustomProcesses.md
```diff
@@ -0,0 +1,302 @@
...
+end
+
+function PointProcesses.integrated_ground_intensity(pp::TwoStateModel, h, a, b)
+    integral = pp.λ * (b - a)
+    interval_events = event_times(h, a, b)
+    if length(interval_events) > 0
+        for i in 1:length(interval_events) - 1
+            integral += pp.Δ * max(interval_events[i + 1] - interval_events[i], pp.τ)
+        end
+        integral += pp.Δ * max(b - interval_events[end], pp.τ)
```
rsenne: I could be wrong but i think this integral is wrong

docs/examples/CustomProcesses.jl
```diff
@@ -0,0 +1,286 @@
...
+end
+
+function PointProcesses.integrated_ground_intensity(pp::TwoStateModel, h, a, b)
+    integral = pp.λ * (b - a)
+    interval_events = event_times(h, a, b)
+    if length(interval_events) > 0
+        for i in 1:(length(interval_events) - 1)
+            integral += pp.Δ * max(interval_events[i + 1] - interval_events[i], pp.τ)
+        end
+        integral += pp.Δ * max(b - interval_events[end], pp.τ)
```
rsenne: See my comment on the .md file

docs/examples/CustomProcesses.jl
```diff
@@ -0,0 +1,286 @@
...
+    TwoStateModel(0.8, 2.0, 0.4, NoMarks()),
+    PoissonProcess(2.0), # Any type of process can be added
+])
+
+sim = simulate(imp, 0.0, 10.0)
+
+print("ground_intensity(imp, 0.0, sim) → $(ground_intensity(imp, 0.0, sim))")
+print("ground_intensity(imp, 0.0, sim, 1) → $(ground_intensity(imp, 0.0, sim, 1))")
+print("intensity(imp, 0.0, 0.0, sim) → $(intensity(imp, 0.0, 0.0, sim))")
+print("intensity(imp, 0.0, nothing, sim, 3) → $(intensity(imp, 0.0, nothing, sim, 3))")
```
rsenne:
```suggestion
print("intensity(imp, 0.0, nothing, sim, 3) → $(intensity(imp, nothing, 0.0, sim, 3))") 
```

?

docs/examples/CustomProcesses.jl
```diff
@@ -0,0 +1,286 @@
...
+    τ::R
+    mark_dist::D
+end
+
+# > Note: Subtype `AbstractUnivariateProcess`, not just `AbstractPointProcess`, this way we can use `TwoStateModel` to build multivariate processes with `IndependentMultivariateProcess`.
+
+# `PointProcessMarkDistribution` encompasses `NoMarks` and any distribution in package
+# `Distributions.jl`. In the next chapter we will discuss custom distributions as well.
+
+# > Even if you plan on only using a non-marked process, you should keep the `mark_dist` field. In you really want you can define an inner constructor like `TwoStateModel(λ, Δ, τ) → new(λ, Δ, τ, Nomarks())`.
```
rsenne:
```suggestion
# > Even if you plan on only using a non-marked process, you should keep the `mark_dist` field. If you really want you can define an inner constructor like `TwoStateModel(λ, Δ, τ) → new(λ, Δ, τ, Nomarks())`.
```

docs/examples/CustomProcesses.jl
```diff
@@ -0,0 +1,286 @@
...
+
+imp = IndependentMultivariateProcess([
+    TwoStateModel(1.2, 2.0, 0.3, Normal()),
+    TwoStateModel(0.8, 2.0, 0.4, NoMarks()),
+    PoissonProcess(2.0), # Any type of process can be added
+])
+
+sim = simulate(imp, 0.0, 10.0)
+
+print("ground_intensity(imp, 0.0, sim) → $(ground_intensity(imp, 0.0, sim))")
```
rsenne:
```suggestion
println("ground_intensity(imp, 0.0, sim) → $(ground_intensity(imp, 0.0, sim))")
```

docs/examples/CustomProcesses.jl
```diff
@@ -0,0 +1,286 @@
...
+print("ground_intensity(imp, 0.0, sim, 1) → $(ground_intensity(imp, 0.0, sim, 1))")
```
rsenne:
```suggestion
println("ground_intensity(imp, 0.0, sim, 1) → $(ground_intensity(imp, 0.0, sim, 1))")
```

docs/examples/CustomProcesses.jl
```diff
@@ -0,0 +1,286 @@
...
+print("intensity(imp, 0.0, 0.0, sim) → $(intensity(imp, 0.0, 0.0, sim))")
```
rsenne:
```suggestion
println("intensity(imp, 0.0, 0.0, sim) → $(intensity(imp, 0.0, 0.0, sim))")
```

rsenne (changes requested):

src/history.jl
JoseKling: Yes. ctrl+c ctrl+v is nice but a bit dangerous ; )
Also added tests for this

src/history.jl
JoseKling: Fixed and updated the docstring. Also replace the `@assert` by proper `DomainError`

src/univariate/poisson/simulation.jl
JoseKling: I kept this one, actually.
With the support for custom mark distributions, the distribution of the marks might depend on the history, including past values of the marks. This implementation supports it, and is closer to `simulate_ogata`.
The only issue is that in the tests for `BootstrapTest`, I was getting an error because of repeated event times. It happened because it used `Float32` precision, and at `t` close to 1000, there was an underflow in `t += rand(rng, inter_dist)` (`Distributions.jl` always returns `Float64` values, so the increment needs to be converted). I guess this happens in `simulate_ogata` as well.
Any ideas for improvement?

rsenne: I'm not sure there is a workaround for this. I think especially for 1000 events, this feels like a fundamental precision issue caused by `Float32`. I guess are options are:

1.) Leave it at the users peril to use F32 ( we could throw a specific warning). I.e., leave tests on F64 and document explicitly that conversion of F64 to F32 is basically guaranteed to cause underflow issues
2.) Add some type of helper that during conversion adds jitter to events to disambiguate

2 is hard to justify albeit likely causing no discernible issues--so I lean towards 1.

rsenne:
> Should we raise a deprecation error or something like that?

I would say until we have evidence for a solid user base for anything like this to just delete as we are pre v1. Anyc hanges we can document (which reminds me, we should keep a changelog)

src/univariate/poisson/simulation.jl
JoseKling: Actually, converting the interarrival right away solved the problem. The problem was constructing the whole vector of event times and then converting in the end.
Anyway, one less problem.

JoseKling:
about the [changelog](https://github.com/JoseKling/PointProcesses.jl/releases). Or do you mean something else?

rsenne:
> Or do you mean something else?

Related idea, but basically we would house a [CHANGELOG.md](https://github.com/depasquale-lab/StateSpaceDynamics.jl/blob/dev/CHANGELOG.md) per the [keep a changelog project](https://keepachangelog.com/en/1.0.0/).

It's basically the same thing but it tends to be more long format. But its not strictly necessary and we can use the same format for releases

JoseKling:
There are a couple of lines not reached by the tests. Apart from that, any other issues here?

rsenne:
Just updated from main--I can take another look

rsenne:
@JoseKling there are 5 unresolved comments from above can you take a look at them and let me know if you agree/disagree? (I think they were missed because they were hidden)

src/abstract_point_process.jl
JoseKling: Hawkes right now is not exactly complete. Like you said, there is no mark_dist for it. But it is going to be fixed very soon. It is actually already done in the PR for multivariate Hawkes processes, just need to check the tests.
About the mark_dist method, not sure we need it (after Hawkes is fixed), but would not hurt, I guess.

rsenne: If this is only a temp problem then it should be fine

docs/src/examples/CustomProcesses.md
JoseKling: Well spotted. I used max instead of min. The period at the higher level is either τ, if the next event occurs after this, or it should account for the whole time between the two events.

src/simulation.jl
JoseKling: you mean not defining a fallback, just raising an error if not explicitly implemented? I have implemented this

```julia
function Base.eltype(md::AbstractMarkDistribution)
    return typeof(sample_mark(md, 0.0, History(0.0, 1.0)))
end
```

rsenne: Ah my bad, I don't think I articulated my concern well (or really at all). Let me try again.

The issue I'm worried about is that this still evaluates the distribution on an empty history at `t=0`, so IIUC anything history or time dependent would blow up. Like if a mark dist does `last(event_marks(h))` it'd throw on the empty history before we ever get a type out

JoseKling: Yes, but the mark_distribution should be defined for empty histories, otherwise we would have a problem starting a simulation. `simulate_ogata` uses this explicitly. If it is dependent on past events, there should be a branch for when `h` is empty, and if is dependent on time, 0 is inside the interval the history is defined, so should be valid.
What is a bit weird about the way things are currently implemented is that we may pass a full history of event times or only a partial history to mark_distribution.
What I mean is, for `simulate_ogata`, the simulation algorithm basically samples the next event after time `t` based on the history up to time `t`, so we would not have to check for event times after `t`. For Hawkes, in contrast, we pass a history with all the event times already simulated (although this can be easily changed). This is related to issue #90.

rsenne: Good point. My example would break anyway given that. 

I guess my remaining worry is narrower:
1. we hardcode `t=0.0` and `History(0.0, 1.0)`. That's only in-domain when the real simulation window includes 0. For a time-dependent dist simulated on, say,` [100, 200]`, we'd be evaluating it at `t=0`
2. `eltype` is a type-level query and shouldn't have to draw a sample (through the rng) to answer it. I'm not really sure what the better way to handle it would be though that would work for both univariate and multivariate cases so maybe i'm overthinking

I'm happy to kick this down the road to fix in #90 or some other issue. The complaint as it stands is pretty minor

rsenne (approved):

src/simulation.jl
JoseKling: 1. Even if 0.0 is not in the simulation interval, in principle the mark_distribution should work for any point process, and would need to be sampleable no matter the interval the history is defined on. Unless it is a really weird distribution in which the type it returns changes with time, it should be ok.
2. To be honest, I am not a fan of this myself, but I figured it would work.

PR #92, opened 2026-06-02
Interface for Multivariate processes

JoseKling:
This PR has the objective of implementing multivariate Hawkes processes.
In the process, some other modifications were carried out.
- Split implementation of univariate Hawkes process into three files and added mark distributions
- Defined separate common interfaces for `AbstractUnivariateProcess` and `AbstractMultivariateProcess`


JoseKling:
I figured it would be a good idea to split this PR in two.  
This part only fixes the univariate Hawkes process and defines the interface for multivariate processes.
The next PR will be the implementation of multivariate Hawkes processes.

rsenne:
Is this ready for me to review?

JoseKling:
Sorry. I guess I forgot to request the review. But it should be ready, yes

src/abstract_point_process.jl
```diff
@@ -122,13 +101,7 @@ Compute the log probability density function for a temporal point process `pp` a
...
 The default method uses a loop over events combined with `integrated_ground_intensity`, but it should be reimplemented for specific processes if faster computation is possible.
 """
-function DensityInterface.logdensityof(pp::AbstractPointProcess, h::History)
-    l = -integrated_ground_intensity(pp, h, min_time(h), max_time(h))
-    for (t, m) in zip(event_times(h), event_marks(h))
-        l += log_intensity(pp, m, t, h)
-    end
-    return l
-end
+DensityInterface.logdensityof
```
rsenne: ?

src/univariate/hawkes/hawkes_process.jl
```diff
@@ -22,136 +22,55 @@ Conditions:
...
-        ψ = D / N
-        ω = D / div
-5. If convergence criterion is met, return updated parameters, otherwise, back to step 2.
-
-Notice that, in the implementation, the process is normalized so the average inter-event time is equal to 1 and, 
-therefore, the interval of the process is transformed from T to N. Also, in equation (8) in the paper,
+HawkesProcess(μ, α, ω) = HawkesProcess(μ, α, ω, NoMarks())
 
-∑_{i=1:n} pᵢᵢ = ∑_{i=1:n} (1 - ∑_{j < i} Dᵢⱼ) = N - D.
+function ground_intensity(hp::HawkesProcess, h::History, t)
```
rsenne: This will throw on a Univariate hawkes no? as in univariate_process.jl the order is t, h?

rsenne: if yes we should also add a test for a univariate hawkes intensity

src/univariate/hawkes/simulation.jl
```diff
@@ -0,0 +1,37 @@
...
+gen_{n+1} is the set of all events simulated from all events in gen_n.
+The algorithm stops when the simulation from one generation results in no further events.
+=#
+function generate_descendants(
+    rng::AbstractRNG, immigrants::Vector{T}, tmax, α, ω
+) where {T<:Real}
+    descendants = T[]
+    next_gen = immigrants
+    while !isempty(next_gen)
+        # OPTIMIZE: Can this be improved by avoiding allocations of `curr_gen` and `next_gen`? Or does the compiler take care of that?
```
rsenne, line 27: probably. we would want to make a workspace type that we use inplace operations on

src/univariate_process.jl
```diff
@@ -0,0 +1,33 @@
...
+    for (t, m) in zip(event_times(h), event_marks(h))
+        l += log_intensity(pp, m, t, h)
+    end
+    return l
+end
+
+function time_change(h::History, pp::AbstractUnivariateProcess)
+    Λ(t) = integrated_ground_intensity(pp, h, min_time(h), t)
+    return time_change(h, Λ)
+end
```
rsenne: redundant with abstract_point_process.jl:126?

src/univariate/hawkes/simulation.jl
```diff
@@ -0,0 +1,37 @@
+function simulate(rng::AbstractRNG, hp::HawkesProcess, tmin, tmax)
+    sim = simulate_poisson_times(rng, hp.μ, tmin, tmax) # Simulate Poisson process with base rate
+    sim_desc = generate_descendants(rng, sim, tmax, hp.α, hp.ω) # Recursively generates descendants from first events
+    append!(sim, sim_desc)
+    sort!(sim)
+    h_temp = History(sim, tmin, tmax)
+    marks = [sample_mark(hp.mark_dist, t, h_temp) for t in event_times(h_temp)]
+    return History(; times=sim, tmin=tmin, tmax=tmax, marks=marks)
```
rsenne: why build history twice?

src/univariate/hawkes/fit.jl
```diff
@@ -0,0 +1,102 @@
...
+"""
+function StatsAPI.fit(
+    ::Type{HawkesProcess{T,MD}},
+    h::History;
+    step_tol::Float64=1e-6,
+    max_iter::Int=1000,
+    rng::AbstractRNG=default_rng(),
+) where {T<:Real,MD<:PointProcessMarkDistribution}
+    n = nb_events(h)
+    n == 0 && return HawkesProcess(zero(T), zero(T), zero(T))
```
rsenne: this will cause type instability

src/multivariate_process.jl
```diff
@@ -0,0 +1,40 @@
+"""
+    AbstractMultivariateProcess
+
+Abstract type for multivariate temporal point processes.
+"""
+abstract type AbstractMultivariateProcess <: AbstractPointProcess end
+
+Base.ndims(pp::AbstractMultivariateProcess) = length(pp.mark_dist)
```
rsenne, line 22: This assumes there is a mark_dist field but IndependentMultivariate does not have one?

src/univariate/hawkes/fit.jl
```diff
@@ -0,0 +1,102 @@
+"""
+    StatsAPI.fit(rng, ::Type{HawkesProcess{T}}, h::History; step_tol::Float64 = 1e-6, max_iter::Int = 1000) where {T<:Real}
```
rsenne: stale docstring

src/HypothesisTests/PPTests/monte_carlo_test.jl
```diff
@@ -110,6 +110,6 @@ function MonteCarloTest(
         throw(ArgumentError("Test is not valid for empty event history."))
     end
 
-    pp_est = fit(PP, h)
+    pp_est = fit(PP, h)::AbstractPointProcess # JET complains if not specified
```
rsenne, line 113: this is bein caused by the type stability issue i'm p sure

src/HypothesisTests/PPTests/bootstrap_test.jl
```diff
@@ -71,7 +71,7 @@ function BootstrapTest(
     end
 
     # Estimate process and calculate statistic from data
-    pp_est = fit(PP, h; rng=rng)
+    pp_est = fit(PP, h; rng=rng)::AbstractPointProcess # JET complains if not specified
```
rsenne: same as my other comment

src/univariate/hawkes/hawkes_process.jl
```diff
@@ -22,136 +22,55 @@ Conditions:
...
-    return fit(HP{T}, h; kwargs...)
+function DensityInterface.logdensityof(hp::HawkesProcess, h::History)
+    A = zeros(nb_events(h)) # Vector A in Ozaki (1979)
+    for i in 2:nb_events(h)
+        A[i] = exp(-hp.ω * (h.times[i] - h.times[i - 1])) * (1 + A[i - 1])
+    end
+    return sum(log.(hp.μ .+ (hp.α .* A))) - # Value of intensity at each event
+           (hp.μ * duration(h)) - # Integral of base rate
+           ((hp.α / hp.ω) * sum(1 .- exp.(-hp.ω .* (duration(h) .- h.times)))) + # Integral of each kernel
+           sum(log.([densityof(hp.mark_dist, t, h, m) for (t, m) in zip(h.times, h.marks)])) # Lok likelihood of marks
```
rsenne:
```suggestion
           sum(log.([densityof(hp.mark_dist, t, h, m) for (t, m) in zip(h.times, h.marks)])) # Loglikelihood of marks
```

rsenne (changes requested):

src/multivariate_process.jl
JoseKling: Yes, `IndependentMultivariateProcess` is a bit different, since we are not really building a new process, but only gluing a bunch of processes together.  
For new multivariate processes, I would expect we use the interface for mark distributions. The idea is that the mark distributions at time `t` are independent conditional on the history up to `t`, so we can model them separetely (well, technically not really true, only for processes in which the probability of simultaneous events is 0, which are all interesting processes I am aware of)

src/univariate/hawkes/hawkes_process.jl
JoseKling: The test is already there, but I switched the order in the tests too : P

src/HypothesisTests/PPTests/monte_carlo_test.jl
JoseKling: This is the message I get.

  ┌ MonteCarloTest(S::Type{<:Statistic}, PP::Type{<:AbstractPointProcess}, h::History; kwargs...) @ PointProcesses /home/jkling/projects/PointProcesses.jl/src/HypothesisTests/PPTests/monte_carlo_test.jl:114
  │ no matching method found `kwcall(::NamedTuple, ::Type{MonteCarloTest}, ::Type{<:Statistic}, ::Rician, ::History)` (1/4 union split): Core.kwcall(merge(NamedTuple()::@NamedTuple{}, kwargs::@Kwargs{…})::NamedTuple, MonteCarloTest, S::Type{<:Statistic}, pp_est::Union{Rician, HawkesProcess, IndependentMultivariateProcess, PoissonProcess}, h::History)
  └────────────────────

Apparently, the static analysis thinks that  `Rician` is a possible return type of `fit` (`Rician` is a `Distributions.Distribution`). Not sure how to deal with this if not by type annotating this line.

src/univariate/hawkes/hawkes_process.jl
```diff
@@ -22,136 +22,55 @@ Conditions:
...
-therefore, the interval of the process is transformed from T to N. Also, in equation (8) in the paper,
-
-∑_{i=1:n} pᵢᵢ = ∑_{i=1:n} (1 - ∑_{j < i} Dᵢⱼ) = N - D.
+function ground_intensity(hp::HawkesProcess, t, h::History)
+    activation = sum(exp.(hp.ω .* (@view h.times[1:(searchsortedfirst(h.times, t) - 1)])))
+    return hp.μ + (hp.α * activation / exp(hp.ω * t))
+end
 
-"""
-function StatsAPI.fit(
```
rsenne, line 79: These fit methods e.g., `fit(HawkesProcess, h)` were delted yet still referenced in docs (see index.md, hawkes.md, and montecarlo.md)

src/univariate/hawkes/simulation.jl
```diff
@@ -0,0 +1,39 @@
+function simulate(rng::AbstractRNG, hp::HawkesProcess, tmin, tmax)
+    sim = simulate_poisson_times(rng, hp.μ, tmin, tmax) # Simulate Poisson process with base rate
+    sim_desc = generate_descendants(rng, sim, tmax, hp.α, hp.ω) # Recursively generates descendants from first events
+    append!(sim, sim_desc)
+    sort!(sim)
+    h = History(eltype(sim)[], tmin, tmax, eltype(hp.mark_dist)[], Nothing[], 1)
+    for event in sim
+        push!(h, event, sample_mark(hp, event, h))
```
rsenne: Currently ignores the RNG

rsenne:
```suggestion
        push!(h, event, sample_mark(rng, hp, event, h))
```

src/univariate_process.jl
```diff
@@ -0,0 +1,47 @@
...
+"""
+    logdensityof(pp, h)
+
+Compute the log probability density function for a temporal point process `pp` applied to history `h`:
+```
+ℓ(h) = Σₖ log λ(tₖ|hₖ) - Λ(h)
+```
+The default method uses a loop over events combined with `integrated_ground_intensity`, but it should be reimplemented for specific processes if faster computation is possible.
+"""
+function DensityInterface.logdensityof(pp::AbstractUnivariateProcess, h::History)
```
rsenne, line 37: With the new type hierarchy, BoundedPointProcess now has no `logdensityofmethod` i.e., this would fail:

```julia
using PointProcesses

h = History([1.0, 2.0, 3.0], 0.0, 10.0)
logdensityof(BoundedPointProcess(PoissonProcess(1.0), 0.0, 10.0), h)
```

src/multivariate_process.jl
```diff
@@ -0,0 +1,70 @@
...
+
+function log_intensity(pp::AbstractMultivariateProcess, m, t, h::History)
+    return [log_intensity(pp, m, t, h, d) for d in 1:ndims(pp)]
+end
+
+function time_change(h::History{T}, pp::AbstractMultivariateProcess) where {T}
+    tmin = typemax(T)
+    tmax = typemin(T)
+    transformed_times = [zeros(T, nb_events(h, d)) for d in 1:ndims(h)]
+    for d in 1:ndims(pp)
```
rsenne, line 70: A 3-dim process with a history whose keyword constructor computed `N=length(unique(dims)) = 2`, which `split_into_chunks` would do, would cause a BBounds error. I  think changiing transformed tiems to index by `1:ndims(pp)` would fix.

src/univariate/hawkes/hawkes_process.jl
```diff
@@ -22,136 +22,55 @@ Conditions:
...
-    return fit(HP{T}, h; kwargs...)
+function DensityInterface.logdensityof(hp::HawkesProcess, h::History)
+    A = zeros(nb_events(h)) # Vector A in Ozaki (1979)
+    for i in 2:nb_events(h)
+        A[i] = exp(-hp.ω * (h.times[i] - h.times[i - 1])) * (1 + A[i - 1])
+    end
+    return sum(log.(hp.μ .+ (hp.α .* A))) - # Value of intensity at each event
+           (hp.μ * duration(h)) - # Integral of base rate
+           ((hp.α / hp.ω) * sum(1 .- exp.(-hp.ω .* (duration(h) .- h.times)))) + # Integral of each kernel
+           sum(log.([densityof(hp.mark_dist, t, h, m) for (t, m) in zip(h.times, h.marks)])) # Loglikelihood of marks
```
rsenne:
```suggestion
           sum(log(densityof(hp.mark_dist, t, h, m)) for (t, m) in zip(h.times, h.marks); init=0.0) # Loglikelihood of marks
```

rsenne: ?

src/bounded_point_process.jl
```diff
@@ -39,3 +39,12 @@ end
...
 end
+
+function time_change(h::History{T}, pp::BoundedPointProcess) where {T}
+    if min_time(h) < min_time(pp) || max_time(h) > max_time(pp)
+        throw(ArgumentError("History is defined outside the bounds of the process"))
+    end
+    return time_change(h, pp.pp)
+end
+
+simulate(pp::BoundedPointProcess) = simulate(pp.pp, pp.tmin, pp.tmax)
```
rsenne, line 57: new fallabck covers this?

src/univariate/poisson/inhomogeneous/inhomogeneous_poisson_process.jl
```diff
@@ -61,9 +61,9 @@ end
 
 ## AbstractPointProcess interface
 
-ground_intensity(pp::InhomogeneousPoissonProcess, t, h) = pp.intensity_function(t)
+ground_intensity(pp::InhomogeneousPoissonProcess, t, h::History) = pp.intensity_function(t)
 
-function intensity(pp::InhomogeneousPoissonProcess, m, t, h)
+function intensity(pp::InhomogeneousPoissonProcess, m, t, h::History)
     return ground_intensity(pp, t, h) * densityof(pp.mark_dist, t, h, m)
 end
```
rsenne: covered by fallback?

docs/examples/CustomProcesses.jl
```diff
@@ -141,15 +141,45 @@ plot(
...
+                upper_bound,
+                init_params,
+                NelderMead(),
+                Optim.Options(; x_reltol=1e-3),
+            )
+
+            optimal = Optim.minimizer(result)
+            solved = true
+
+        catch DomainError
```
rsenne: this does not filter by type. need something like:

```julia

catch e
    if isa(e, DomainError)
        n_tries += 1
    else
        rethrow()
    end
end
```

or similar

rsenne: this is shorter:

```julia
catch e
    e isa DomainError || rethrow()
    n_tries += 1
end
```

src/univariate/poisson/fit.jl
```diff
@@ -8,8 +13,10 @@ function StatsAPI.fit(
     return PoissonProcess(λ, mark_dist)
 end
 
-function StatsAPI.fit(pptype::Type{PoissonProcess{R,D}}, args...; kwargs...) where {R,D}
```
rsenne, line 11: this deletion means we no longer support weighted fitting. is that intentional?

src/univariate/hawkes/hawkes_process.jl
```diff
@@ -22,136 +22,55 @@ Conditions:
...
-    T = promote_type(Float64, H)
-    return fit(HP{T}, h; kwargs...)
+function DensityInterface.logdensityof(hp::HawkesProcess, h::History)
+    A = zeros(nb_events(h)) # Vector A in Ozaki (1979)
+    for i in 2:nb_events(h)
+        A[i] = exp(-hp.ω * (h.times[i] - h.times[i - 1])) * (1 + A[i - 1])
+    end
+    return sum(log.(hp.μ .+ (hp.α .* A))) - # Value of intensity at each event
+           (hp.μ * duration(h)) - # Integral of base rate
+           ((hp.α / hp.ω) * sum(1 .- exp.(-hp.ω .* (duration(h) .- h.times)))) + # Integral of each kernel
```
rsenne: should be `h.tmax .- h.times` ?

src/univariate/hawkes/hawkes_process.jl
```diff
@@ -22,136 +22,55 @@ Conditions:
...
-5. If convergence criterion is met, return updated parameters, otherwise, back to step 2.
-
-Notice that, in the implementation, the process is normalized so the average inter-event time is equal to 1 and, 
-therefore, the interval of the process is transformed from T to N. Also, in equation (8) in the paper,
-
-∑_{i=1:n} pᵢᵢ = ∑_{i=1:n} (1 - ∑_{j < i} Dᵢⱼ) = N - D.
+function ground_intensity(hp::HawkesProcess, t, h::History)
+    activation = sum(exp.(hp.ω .* (@view h.times[1:(searchsortedfirst(h.times, t) - 1)])))
+    return hp.μ + (hp.α * activation / exp(hp.ω * t))
+end
```
rsenne, line 51: need the shift invariant form

src/history.jl
```diff
@@ -153,8 +153,13 @@ end
 function History(; times, tmin, tmax, marks=nothing, dims=nothing, check_args=true)
     if times isa Vector{<:Real}
         marks === nothing && (marks = fill(nothing, length(times)))
-        dims === nothing && (dims = fill(nothing, length(times)))
-        return History(times, tmin, tmax, marks, dims; check_args=check_args)
+        if dims === nothing
+            dims = fill(nothing, length(times))
+            N = 1
+        else
+            N = length(unique(dims))
```
rsenne: This counts distinct dim values present, not the actual dimension count

src/univariate/hawkes/hawkes_process.jl
```diff
@@ -22,136 +22,55 @@ Conditions:
...
+    mark_dist::D
 
-    function HawkesProcess(μ::T1, α::T2, ω::T3) where {T1,T2,T3}
+    function HawkesProcess(
+        μ::T1, α::T2, ω::T3, mark_dist::D
+    ) where {T1,T2,T3<:Real,D<:PointProcessMarkDistribution}
         any((μ, α, ω) .< 0) &&
             throw(DomainError((μ, α, ω), "All parameters must be non-negative."))
         (α > 0 && α >= ω) &&
             throw(DomainError((α, ω), "Parameter ω must be strictly smaller than α"))
```
rsenne: error message inverted?

rsenne (changes requested):

rsenne:
found a few more bugs. i'll see if i can figure out the JET thing

src/univariate_process.jl
JoseKling: How useful is the `BoundedPointProcess`? I don't really see any need for it, honestly...

src/multivariate_process.jl
JoseKling: but should we allow for `time_change` to be applied with process and history with different dimensions? I think it is better for `split_into_chunks` to guarantee all the chunks have the same number of dimensions as the original history.

src/bounded_point_process.jl
JoseKling: Well, this only delegates `simulate` to the method defined for `pp.pp`. If the inner process `pp.pp` has `simulate` defined or, at least, `ground_intensity` and `ground_intensity_bound`, it works.

JoseKling: ah, I see, the question is if this is now redundant.  
Not really, since `BoudedPointProcess` is not really a process, is just a 'container' for a process. As I mentioned in some other comment, I think this does not add much benefit.

src/history.jl
JoseKling: yes. In the case we are passing vectors, instead of one vector per dimension, this is the only way to deduce the number of dimensions in the history.  
I agree this has some weird edge cases. We should probably think in a more structured way about the interface for the constructor. Perhaps this is trying to be too general.

src/univariate_process.jl
rsenne: I'm happy to get rid of it--want to do it in a separate PR though? Then this can be rebased off that one

src/history.jl
rsenne: some ideas:

- Deduce with `N = maximum(dims)` instead of `length(unique(dims))`
- Additionally accept an explicit `N/ndims` keyword so callers like `split_into_chunks` can preserve the parent's dimension count

src/multivariate_process.jl
rsenne: sure, that makes sense. In that case we probably want to thread N through:

`chunk = History(times, a, b, marks, dims, h.N; check_args=false)`

src/history.jl
JoseKling: In the `split_into_chunks` I planned to do exactly this.
About deducing the dimensions with `max(dims)`, this also has edge cases in which there are no events in the dimensions with higher values. Perhaps for multivariate histories the dimension must be passed, if the number of dimensions is nos passed we assume it is univariate?

src/univariate_process.jl
JoseKling: Sure, let's leave for the next PR.

src/univariate/poisson/fit.jl
JoseKling: I think I was testing to see if this was the origin of the problem with JET and forgot to add back. Anyway, good catch

JoseKling: Even though this is not really in tune with the rest of the package anymore. Not sure if we should keep it, or try to implement something similar for the other processes, or get rid of it...

JoseKling:
I think I got everything, except for the JET warning.
I changed History slightly. Instead of guessing the number of dimensions, if a Vector is passed to the constructor, we assume it is a univariate history. The inner constructor does not have any default parameters, so everything must be passed explicitly.
Also, I changed the constructor for the empty history to expect a type for the marks and the number of dimensions (not necessary for univariate history)

rsenne:
i'll try to look at this soon

rsenne (approved):

rsenne:
See #96

PR #95, opened 2026-06-16
Pin JuliaFormatter to v1.0.62

### LINKED ISSUE #94: Pin JuliaFormatter version
The JuliaFormatter action is really noisy imo (I know i introduced it, in hindsight this was a mistake). I think given the [recommendation of the package](https://github.com/julia-actions/julia-format/pull/51#issuecomment-4642607713) we should pin JuliaFormatter and reintroduce it as part of testing. That way each PR isn't being bombarded by 30 commits everytime someone forgets to run `format(".")`.

rsenne:
This PR pins JuliaFormatter to 1.0.62. This has a few advantages.

1. v1.0.62 is considered stable by the package and so is reliable in comparison to v2 where format changes are not unexpected minor version to minor version
2. We can reintroduce testing w/ JET as that package loosened its JuliaSyntax comapt to v1 (i.e., no more `MethodTableError`)
3.  We can get rid of the format on PR stuff that imo just was extremely noisy

Resolves #94


JoseKling (approved):

JoseKling:
Great. Thanks

PR #96, opened 2026-08-08
Review Changes

rsenne:
hey @JoseKling i found a few more things on #92 during review but I don;t want to keep holding it up so I'm doing a stacked PR here with some proposed changes. Sorry this has taken me so long. I think these all make sense but let me know if anything is afoot. I've approved #92 so this can be the next PR after

Bug fixes:
- `simulate(::PoissonProcess, tmin, tmax)` started its first inter-arrival at 0
  instead of `tmin`, so simulating on an interval with `tmin != 0` produced events
  outside `[tmin, tmax)` and a positive log-density.
- `simulate(::PoissonProcess, ...)` and `simulate_ogata` sampled marks through the
  rng-less `sample_mark`, so seeded simulations were not reproducible in their marks.
- `event_times(h, d)` ignored `d` for univariate histories while `event_marks(h, d)`
  did not. `History(h, d)` combines both and passes `check_args=false`, so
  `History(univariate_h, 2)` silently returned a history whose times and marks had
  different lengths. Both accessors now agree, and `History(h, d)` validates `d`.
- The inner `History` constructor coerced `dims .= nothing` when `N == 1`, which
  throws a `convert` `MethodError` for integer dims and mutates the caller's vector
  for `Union{Nothing,Int}` dims. It now validates instead of coercing.
- The generic multivariate `time_change` sized its buffers by `ndims(h)` but looped
  over `1:ndims(pp)`, giving a `BoundsError` when the process had more dimensions
  than the history. Mismatched dimensions now throw a `DimensionMismatch`.

JET:
- `fit(::Type{<:Distribution}, ::History)` and `fit(::Type{<:AbstractPointProcess},
  ::History)` intersect at `Type{Union{}}`, so inference put a `Distribution` in the
  return-type union of `fit(PP, h)`, which is what JET was reporting. Adding a
  `Type{Union{}}` method removes it, so the `::AbstractPointProcess` assertion in
  `MonteCarloTest` (which turned the inference gap into a possible runtime
  `TypeError`) is no longer needed.

Docs:
- The multivariate interface docstring listed `integrated_ground_intensity(pp, t, h, d)`
  and required `simulate(pp, h)`
- The interface docstrings in `abstract_point_process.jl` documented the multivariate
  methods as `(pp, h, t, d)`, but the code takes `(pp, t, h, d)`.
- `mark_distribution` was declared and documented twice, so the first docstring was
  discarded with a `Replacing docs` warning during precompilation. Merged into one.
- Restored the `time_change(h, Λ)` docstring, without which `@docs time_change` in
  `api.md` rendered the `InhomogeneousPoissonProcess` method under History.
- Documented the current `History` constructor forms, including that a plain vector
  of times always builds a univariate history.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /docs/build/
 **/Manifest.toml
 .vscode
+dev/
--- a/src/HypothesisTests/PPTests/monte_carlo_test.jl
+++ b/src/HypothesisTests/PPTests/monte_carlo_test.jl
@@ -110,6 +110,6 @@ function MonteCarloTest(
         throw(ArgumentError("Test is not valid for empty event history."))
     end
 
-    pp_est = fit(PP, h)::AbstractPointProcess # JET complains if not specified
+    pp_est = fit(PP, h)
     return MonteCarloTest(S, pp_est, h; kwargs...)
 end
--- a/src/abstract_point_process.jl
+++ b/src/abstract_point_process.jl
@@ -16,12 +16,13 @@ Base.ndims(::AbstractPointProcess)
 
 ## Intensity functions
 """
-    ground_intensity(pp, h, t)
+    ground_intensity(pp, t, h)
+    ground_intensity(pp, t, h, d)
 
 Compute the ground intensity for a temporal point process `pp` applied to history `h` at time `t`.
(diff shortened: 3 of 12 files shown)


JoseKling (approved):
Great, thanks!

